PR #15, opened 2017-04-26
Change templates to not auto play and preload videos.

godane:


kelson42:
@godane Why autoplay is a bad idea?

dattaz:
Hi,
Thanks for your contribution !
The idea of this scraper is to have the same behavior that on youtube, so i'm not understanding why you want to make these changes.

godane:
We could change the code to have autoplay as a option. I'm thinking autoplay is bad on wifi/mesh network so disabling it is a good idea.

kelson42:
@godane Why this is bad?

godane:
To me videos should only play when i'm clicking on there windows. Not just when i load the page. I also want to only load the files that i'm clicking on. Video files are too big some times and i don't want them downloading in the background. 

The only middle ground i see is disable it on the welcome page. 

PS i noticed that videos are not in the same order has the youtube channel/playlist. Mostly cause when i download the Gaming Historian channel the welcome trailer was not the same and the videos list are not in the same order at all.

dattaz:
I understand your point of view, but as says before the goal is to have the same behavior than youtube.
For your PS i have open an new issue #16

PR #26, opened 2018-06-22
Make MP4 streamable #25

### LINKED ISSUE #25: MP4 encoded video are not streamable
They have to be fully downloaded to be playable.

to avoid this the following ffmpeg options should be used: `-movflags +faststart`

It is unclear how it works for avconv

kelson42:
Fix #25


dattaz:
Thanks !

PR #43, opened 2019-09-16
New version based on Youtube Data API v3

rgaudin:
Here's the new version of the scraper.

It started as a fix of the current version which is based on Youtube's DOM making it quite fragile.
Code has been refactored to make it easier to maintain over time and harmonize with our other projects (docker, etc).

Please take a look at the changelog for details of what has changed.

Current focus in on creating good Youtube ZIMs that work on every reader (we're fixing them one by one) but looking forward, we might want to defer that API-specifics to youtube-dl to support additional video platforms (and not be tied to the API quotas). Hopefully, it would be easier with this version.

@dattaz, very sorry I dropped your last commit (implementing #32) as this was already taken care of (that's why I assigned myself 🤓) and the amount of refactoring made it very difficult to rebase your last commit-in.


dattaz:
@rgaudin no problem ! Thank you for all the work done !

PR #57, opened 2019-11-27
Add a CONTRIBUTING.md file

kelson42:
@rgaudin It is only a stub, a few things are wrong or incomplete, would be nice if you could fix it.


rgaudin:
been taken care of in other PR

PR #67, opened 2020-02-24
Add feature to filter videos by date

### LINKED ISSUE #31: Sub select videos by date
I would like to be able to run youtube2zim and only gather newer that a certain date.
This would allow splitting a youtube channel zim into smaller parts. Also it would allow creating a new zim for newer content only, without having to modify already downloaded youtube channel zims.

deborahlow97:
Fixes #31 

### Context
- Allow user to add a parameter to filter and download videos uploaded on or after the specified date.

### Approach
- Used `youtube_dl.DateRange` to specify daterange

### Case: video upload date isn't within range: (Date specified: 20191212)
image

### DIFF
--- a/youtube2zim/entrypoint.py
+++ b/youtube2zim/entrypoint.py
@@ -167,6 +167,10 @@ def main():
         action="version",
         version=SCRAPER,
     )
+    parser.add_argument(
+        "--dateafter",
+        help="Custom filter to download videos uploaded on or after specified date. Format: YYYYMMDD or (now|today)[+-][0-9](day|week|month|year)(s)?"
+    )
 
     args = parser.parse_args()
     logger.setLevel(logging.DEBUG if args.debug else logging.INFO)
--- a/youtube2zim/scraper.py
+++ b/youtube2zim/scraper.py
@@ -65,6 +65,7 @@ class Youtube2Zim(object):
         language,
         locale_name,
         tags,
+        dateafter,
         title=None,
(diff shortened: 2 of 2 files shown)


kelson42:
@deborahlow97 Thank you very much for your patch. This will be reviewed around the end of the week, so please don't wonder if you see no activity in the next days on this PR. What would be really helpful, if you want to continue with Python, would be if you could have a look to bugs on Sotoki https://github.com/openzim/sotoki.

deborahlow97:
> @deborahlow97 Thank you very much for your patch. This will be reviewed around the end of the week, so please don't wonder if you see no activity in the next days on this PR. What would be really helpful, if you want to continue with Python, would be if you could have a look to bugs on Sotoki https://github.com/openzim/sotoki.

Thank you for the speedy reply! Am happy to take a look at the issues at Sotoki 😀

deborahlow97:
> Thanks @deborahlow97 !
> 
> Unfortunately, this can't work as-is as Youtube-dl is used for downloading videos only but not for selecting/filtering them. This is done via the Youtube API, currently only using the `--type` and `--id` info.
> 
> what you could do it keep the `youtube_dl.DateRange` which a good human-to-range parser and use those dates to change API calls so we'd only get filtered results. There should be no need for the `daterange` YoutubeDL param afterward.
> 
> Let me know if I can be of any additional help

Thanks for pointing that out! I'll fix that.

PR #68, opened 2020-02-27
Add feature to filter videos by date

### LINKED ISSUE #31: Sub select videos by date

deborahlow97:
Fixes #31 

### Context
- Allow user to add a parameter to filter and download videos uploaded on or after the specified date.

### Approach
- Used `youtube_dl.DateRange` to specify daterange
- Used Youtube API to get published date

### DIFF
--- a/youtube2zim/scraper.py
+++ b/youtube2zim/scraper.py
@@ -81,7 +81,7 @@ class Youtube2Zim(object):
         self.collection_type = collection_type
         self.youtube_id = youtube_id
         self.api_key = api_key
-        self.dateafter = youtube_dl.DateRange(dateafter)
+        self.dateafter = dateafter
 
         # video-encoding info
         self.video_format = video_format
@@ -228,11 +228,13 @@ class Youtube2Zim(object):
         )
 
     def run(self):
+        """ validate dateafter input """
+        self.validate_dateafter_input()
+
         """ execute the scraper step by step """
         logger.info(
             f"starting youtube scraper for {self.collection_type}#{self.youtube_id}"
         )
-
         logger.info("preparing build folder at {}".format(self.build_dir.resolve()))
         if not self.keep_build_dir and self.build_dir.exists():
             shutil.rmtree(self.cache_dir, ignore_errors=True)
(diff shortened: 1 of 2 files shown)


rgaudin:
Great! It's working fine. Couple things I'd love though:

* `DateRange` is very permissive but can still fail with garbage value. Could you protect that and make sure user gets an informative message in case of incorrect value? I'd prefer that to happen in `run()` so you could just store it in `__init__` and convert/fail to `DateRange` in `run`
* Logs should display the daterange as well ; in the same way we display the nb of playlists/videos.
* in `filter_videos_not_within_date_range`, I see no need for your `date_range` variable. Can we get rid of it? Actually, I think you used it for sementic purposes and that makes me think we should maybe store the date range in a `self.date_range` instead. The parameter can stay as it is, as it implies youtube-dl compatibility.

deborahlow97:
> Great! It's working fine. Couple things I'd love though:
> 
> * `DateRange` is very permissive but can still fail with garbage value. Could you protect that and make sure user gets an informative message in case of incorrect value? I'd prefer that to happen in `run()` so you could just store it in `__init__` and convert/fail to `DateRange` in `run`
> * Logs should display the daterange as well ; in the same way we display the nb of playlists/videos.
> * in `filter_videos_not_within_date_range`, I see no need for your `date_range` variable. Can we get rid of it? Actually, I think you used it for sementic purposes and that makes me think we should maybe store the date range in a `self.date_range` instead. The parameter can stay as it is, as it implies youtube-dl compatibility.

👍 Alright, will do! Thanks for the informative feedback 😄

youtube2zim/scraper.py
```diff
@@ -387,7 +400,7 @@ def extract_playlists(self):
         """ prepare a list of Playlist from user request
 
             USER: we fetch the hidden channel associate to it
-            CHANNEL (and USER): we grab all playlists + `uploads` playlist
+            CHANNEL (and USER): we grab all playlistFs + `uploads` playlist
```
rgaudin: typo here?

deborahlow97:
Currently, if no `--dateafter` is provided, `INFO:videos in date range: 0001-01-01 - 9999-12-31` would be logged.

rgaudin:
Thank you @deborahlow97 !

PR #79, opened 2020-04-14
Fixed #69: introduce optimization cache

### LINKED ISSUE #69: Introduce optimization cache
Like in MWoffliner, see https://github.com/openzim/mwoffliner/issues/996.

If the option to deal with S3 should be availbel at `youtube2zim` level the implementation should be made generic at `python_scraperlib` level so we can reuse it for an other project.

nabinkhadka:
Issue #69


nabinkhadka:
@rgaudin Yes, there are many things to work on and that is the reason I said "working on issue #69" in the description above.  And so I hadn't created a PR and wanted to check with you to see if I am going right direction. But thank you for your time and comments.

rgaudin:
Yes, all of this is fixable. Hit the re-request button when you're ready.

nabinkhadka:
@rgaudin What is a non-optimized version? Is it before running `post_process_video` function which runs the `recompress_video`? If so why would we save a non-optimized version?

rgaudin:
>     * What is a non-optimized version? Is it the one before running `post_process_video` function which runs the `recompress_video`? If so why would we save a non-optimized version?

Yes that's the one. We ask google for the _best_ video it can give us in the format requested but don't ask for a specific quality because those are not reliably available on videos. Instead we get that video file and we recompress it anyway (with different options depending on `--low-quality`). post-processing also includes thumbnail resizing I think.
We want to cache that compressed version so that, if present in cache, we save on both bandwith and processing time/resources.
In your PR, you upload (and thus cache) the received video before processing it. We want to upload the optimized version (hence the parameter name).

I'll use this opportunity to elaborate on the other requirement: `--use-any-optimized-version`.
This processing/reencoding code, if changed, has a lot of impact on the video : they can look better/worst or filesize can change drastically.
So we want to introduce a _version string_ that defines the encoder code. It's just something that we change whenever we make significant changes to the encoding code.

So we want to also store that encoder-version on all the files we upload to the cache (as a meta). This would allow us to programmaticly (not in `youtube2zim`) remove older version for instance.
In reality, we may have to keep running the old-encoder-version-video files because we don't have the time or capacity to re-download from youtube, re-encode and re-upload.
So in this case, we use `--use-any-optimized-version` to say “if there's a cache for this file, no matter the encoder-version, give it to me”.
Otherwise (default), you check for the existence of a video in cache, according to the encoder-version.
The kiwixstorage lib already supports this (see `has_object_matching_meta()`).

Let me know if this is clearer now. it's a lot of details I know :)

>     * Why do we need to know if anything has failed except for logging message? If the download  fails, it's safe to download using youtube-dl as per my perspective. And if the upload fails, we simply don't do it. There is no point in counting how many cache download/upload failed because the user will eventually look for result no matter how they get.

maybe some user doing this just once doesn't care ; but we are not a regular user. we create **a lot** of ZIM files containing **a lot** (thousands) of videos… every month.
It is critical that we know when something that is supposed to work doesn't; especially on retrieving and saving those videos because failure there means that we'll be transferring off youtube, exposing ourselves to IP restrictions, jeopardizing all the ZIM files.

So it's very important. And we want details (exception trace). Out of scope for you but we may want to crash on such scenarios in the future.

>     * I have used python black for formatting and docstrings are also present. Is there some other standards to follow regarding these?

Yes, the failing black file is not the one you worked one. My bad. Thank you for picking that up, it's much appreciated 👍 

As for docstring, there's no written rule but you'll notice they all look somewhat similar and have a similar wording style. Here's an example:

``` py
    # that's yours
    def s3_cache_found(self, video_id, local_video_path):
        """
        Download file if present
        return True if cache is found
        else return False
        """
        pass

    # what I'd prefer to have
    def s3_cache_found(self, video_id, local_video_path):
        """ whether the file was found (and downloaded from) in cache """ 
        pass
```

Hope that's understandable. If not, we'll fix those together once the rest of done.

rgaudin:
> * `zimscraperlib` and `kiwixstorage` have conflict regarding `requests`, so the fix needs to be on `kiwixstorage` rather. I believe this will help build the docker image. Also, I have removed `requests` from the `requirements.txt` file in recent commit

I think we need to change both. requests uses semantic versioning as well so I'll update kiwixstorage to accept `>=2.23,<3.0`. We should also update here and the same version string would work fine for us as well.
You can put git master in requirements for kiwixstorage while you develop. We'll make a new release of kiwixstorage once we've fixed everything you need.

> * Right now there are two options while recompressing, viz libvpx and h264 for video and libvorbis and aacfor for audio.  So depending upon the `--low-quality` argument either of these combine together and will be used for encoding. Hence there can be two values for `encoder_version` meta on each object. Is that correct? Or we save encoding information of audio and video information separately in meta?

No we don't need that granularity. That would be a nightmare to manage. Just a simple version string to match the _code version_ that was used to encode the video.

> * If encoder_version is simply strings like v1, v2, etc, who defines what v1 is, what v2 is, etc?

the developer making changes to the ffmpeg params. you will have to add `v1` for current version and that will stay like that until we change ffmpeg params in a way that has significant impact on the video files. 

> * And in case we introduce new encodings like h999  instead of h264, then the `encoder_version` can have 3 different values?

Nope, as I said, if we change the codec, we'll just bump the version for all.

> * `if there's a cache for this file, no matter the encoder-version, give it to me` means we look for `video_format/*/videoId` where `*` can either be  `low` or `high`, correct?  Or just ignore what is in meta?
 
No, just don't filter using the meta but the key remains the same (`{video_format}/{quality}/{videoId}`)

> * `So it's very important. And we want details (exception trace). Out of scope for you but we may want to crash on such scenarios in the future.` I have added error log. But if needed, I can raise an error. If there is proper exceptions (like ClientError, NetworkError, etc) defined in the `kiwixstorage` package, it would be the best way of doing it. What else could be done?

I think error log is good enough for now.

>     * If some video is found in cache, I don't think we need to recompress it as it already has been compressed and saved on S3 storage. Or there could be any other scenario?

No, if downloaded from cache, no need to recompress. that's the point. That's why it's important you upload the compressed version.

> Even though I have asked these questions, I am happy that I can start working on `--use-any-optimized-version` argument.

Good. Hopefully answers above regarding the encoder version should allow you to.

nabinkhadka:
Any suggestion what to do for this line? https://github.com/openzim/youtube/blob/58748bee394c61d90a432b1a44d91cd8f1e3def4/youtube2zim/scraper.py#L288

This won't be true if cache is found, but this line comes before checking the cache presence.

rgaudin:
> Any suggestion what to do for this line?
> 
> https://github.com/openzim/youtube/blob/58748bee394c61d90a432b1a44d91cd8f1e3def4/youtube2zim/scraper.py#L288
> 
> This won't be true if cache is found, but this line comes before checking the cache presence.

Change it to *Quality: regular|low* or something. That's mostly useful to know in the logs what was requested.

nabinkhadka:
Forked the kiwixstorage lib and made dependency version fix [here](https://github.com/nabinkhadka/python_storagelib). Using this docker image was build using minor changes (see attachment
Screen Shot 2020-04-16 at 19 23 54
). Tested my arguments, works pretty good. I have made the PR ready for review.

rgaudin:
Why did you fork it? I've fix the reqs in the lib when I commented and asked you to you its master and bump our requests to `>=2.23,<3.0`. Can you please do that while I review the rest?

youtube2zim/scraper.py
```diff
@@ -140,6 +151,17 @@ def __init__(
             build_dir=self.build_dir, api_key=self.api_key, cache_dir=self.cache_dir
         )
 
+        # Optimization-cache
+        self.s3_url_with_credentials = s3_url_with_credentials
+        self.use_any_optimized_version = use_any_optimized_version
+        self.video_quality = "low" if self.low_quality else "high"
+        if s3_url_with_credentials:
+            self.s3_storage = KiwixStorage(s3_url_with_credentials)
```
rgaudin, line 159: please init that with `None` in `__init__`

youtube2zim/scraper.py
```diff
@@ -140,6 +151,17 @@ def __init__(
...
+            if not self.s3_storage.check_credentials(list_buckets=True):
```
rgaudin, line 160: why not using `failsafe=True`? This may crash

youtube2zim/scraper.py
```diff
@@ -140,6 +151,17 @@ def __init__(
...
+                raise AuthenticationError(
```
rgaudin, line 161: Why use this Exception? Since its failed most probably from user provided input, use `ValueError`.

youtube2zim/scraper.py
```diff
@@ -140,6 +151,17 @@ def __init__(
...
+                    f"An error occurred when making connection with the s3. Check your key"
+                )
+            self.bucket_name = self.s3_storage.bucket_name
```
rgaudin, line 164: No need to store the bucket name

youtube2zim/scraper.py
```diff
@@ -285,7 +307,7 @@ def run(self):
             f"downloading all videos, subtitles and thumbnails (concurrency={self.max_concurrency})"
         )
         logger.info(f"  format: {self.video_format}")
-        logger.info(f"  recompress: {self.low_quality}")
+        logger.info(f"  quality: {self.video_quality}")
```
rgaudin, line 310: you must add log informing that you'll use the cache.
Also indicate the netloc and the bucket name.

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
 
         return overall_succeeded, overall_failed
 
-    @staticmethod
-    def download_video_files_batch(options, videos_ids):
+    def download_s3_cache(self, s3_key, local_video_path):
```
rgaudin, line 575: rename to `download_from_cache` and use `""" whether it successfully downloaded from cache """`

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+            self.s3_storage.download_file(s3_key, local_video_path)
+        except Exception as ex:
+            logger.error(f"{self.bucket_name}/{s3_key} download failed with {str(ex)}")
+            return False
+        logger.info(
+            f"cache ({self.bucket_name}/{s3_key}) downloaded {local_video_path}"
+        )
+        return True
 
+    def upload_s3_cache(self, s3_key, local_video_path):
```
rgaudin, line 599: rename to `upload_to_cache` and use `""" whether it successfully uploaded to cache """`

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+        try:
+            if self.use_any_optimized_version:
+                if not self.s3_storage.has_object(s3_key, self.bucket_name):
+                    return False
+            else:
+                if not self.s3_storage.has_object_matching_meta(
+                    s3_key,
+                    tag="encoding_version",
+                    value=ENCODER_VERSION,
+                    bucket_name=self.bucket_name,
```
rgaudin, line 586: No need to pass `bucket_name` here

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+                ):
+                    return False
+            os.makedirs(os.path.dirname(local_video_path), exist_ok=True)
```
rgaudin, line 589: use `pathlib`

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
 
         return overall_succeeded, overall_failed
 
-    @staticmethod
-    def download_video_files_batch(options, videos_ids):
+    def download_s3_cache(self, s3_key, local_video_path):
+        """returns True if successfully downloaded cache, otherwise False"""
+        try:
```
rgaudin, line 577: only the download should be inside the try block.

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+                    s3_key,
+                    tag="encoding_version",
+                    value=ENCODER_VERSION,
+                    bucket_name=self.bucket_name,
+                ):
+                    return False
+            os.makedirs(os.path.dirname(local_video_path), exist_ok=True)
+            self.s3_storage.download_file(s3_key, local_video_path)
+        except Exception as ex:
+            logger.error(f"{self.bucket_name}/{s3_key} download failed with {str(ex)}")
```
rgaudin, line 592: remove bucket_name from log. don't use `str()`, it's already in an f-string.

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+            return False
+        logger.info(
+            f"cache ({self.bucket_name}/{s3_key}) downloaded {local_video_path}"
```
rgaudin, line 595: remove bucket_name from log.

youtube2zim/scraper.py
rgaudin, line 575: use `key`, and `video_path` for param names

youtube2zim/scraper.py
rgaudin, line 599: use `key`, and `video_path` for param names

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+    def download_s3_cache(self, s3_key, local_video_path):
+        """returns True if successfully downloaded cache, otherwise False"""
+        try:
+            if self.use_any_optimized_version:
+                if not self.s3_storage.has_object(s3_key, self.bucket_name):
+                    return False
+            else:
+                if not self.s3_storage.has_object_matching_meta(
+                    s3_key,
+                    tag="encoding_version",
```
rgaudin, line 584: it's `encoder_version` not encoding_version. Fix in upload as well.

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+        )
+        return True
 
+    def upload_s3_cache(self, s3_key, local_video_path):
+        """returns True if video uploaded to S3 for cache otherwise None"""
+        try:
+            self.s3_storage.upload_file(
+                local_video_path, s3_key, meta={"encoding_version": ENCODER_VERSION}
+            )
+            logger.info(f"cached video to {self.bucket_name} bucket ({s3_key})")
```
rgaudin, line 605: remove bucket name from log

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+            return True
+        except Exception as ex:
+            logger.error(f"{self.bucket_name}/{s3_key} upload failed with {str(ex)}")
```
rgaudin, line 608: same as above

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+
+    def download_video_files_batch(self, options, videos_ids):
         succeeded = []
         failed = []
-        with youtube_dl.YoutubeDL(options) as ydl:
-            for video_id in videos_ids:
-                try:
+        for video_id in videos_ids:
+            video_location = options["y2z_videos_dir"].joinpath(video_id)
+            # Reset the skip_download argument
```
rgaudin, line 615: asked you to add a comment for this (because option name is not explicit: doesn't downloads the video but downloads the rest)

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+            options["skip_download"] = False
```
rgaudin, line 616: I don't like this. dict are passed by reference ; pretty sure this will blow up when using concurrency

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+            cache_present = None
+            if self.s3_url_with_credentials:
```
rgaudin, line 618: don't test this, test `self.s3_storage`

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+                s3_key = f"{self.video_format}/{self.video_quality}/{video_id}"
+                video_path = os.path.join(video_location, f"video.{self.video_format}")
```
rgaudin, line 620: use `pathlib`!

youtube2zim/scraper.py
```diff
@@ -550,24 +572,71 @@ def get_slot():
...
+                    video_location,
+                    video_id,
+                    options["y2z_video_format"],
+                    options["y2z_low_quality"],
+                    recompress=not cache_present,
+                )
+                succeeded.append(video_id)
+            except (youtube_dl.utils.DownloadError, FileNotFoundError):
+                failed.append(video_id)
+            if cache_present is False:
```
rgaudin, line 637: usage is correct but it's not explicit.  you're testing `cache_present` even if not using s3. Change variable name and/or make condition more explicit.

youtube2zim/scraper.py
rgaudin, line 608: why return `None`? Stay consistent and return `False`

requirements.txt
```diff
@@ -1,8 +1,8 @@
 python-slugify==3.0.3
-requests==2.22.0
 # youtube-dl should be updated as frequently as possible
 youtube_dl
 python-dateutil==2.8.0
 jinja2==2.10.1
 babel==2.7.0
 zimscraperlib>=1.0.1,<1.1
+kiwixstorage==0.1
```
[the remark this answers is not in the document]
nabinkhadka: You mean this in requirements.txt?
`-e git://github.com/kiwix/python_storagelib.git#egg=python_storagelib`

youtube2zim/scraper.py
nabinkhadka: That's why I added comment below it. Why should we add more cost even though we have comment already for the same purpose?

youtube2zim/scraper.py
nabinkhadka: I don't understand what you mean. Though I have a comment few statements below.

requirements.txt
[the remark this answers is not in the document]
rgaudin: yes, that's until we're sure you don't need anymore fixes in the lib and make a release.

youtube2zim/scraper.py
rgaudin: the youtube-dl option is not self explaintatory. One not familiar with it might think it doesn't download anything so please explain what this option does (skip video dl but downloads the rest)

youtube2zim/scraper.py
rgaudin: I see the comment. Please use a better variable name and get rid of the comment.

youtube2zim/scraper.py
nabinkhadka: I suggest not doing this here, because we need to recheck few things here

rgaudin: what things? can you elaborate ?

nabinkhadka: NVM

youtube2zim/scraper.py
nabinkhadka: Planning to send a copy.

rgaudin: Yes, unfortunately I don't think we can avoid it.

youtube2zim/scraper.py
```diff
@@ -572,71 +578,69 @@ def get_slot():
...
-                    value=ENCODER_VERSION,
-                    bucket_name=self.bucket_name,
-                ):
-                    return False
-            os.makedirs(os.path.dirname(local_video_path), exist_ok=True)
-            self.s3_storage.download_file(s3_key, local_video_path)
+            self.s3_storage.download_file(key, video_path)
+            logger.info(f"cache ({key}) downloaded {video_path}")
+            return True
         except Exception as ex:
```
rgaudin, line 596: could you use `exc` instead of `ex` for the exception variable? Not important but that's a convention we use. applies to upload as well

youtube2zim/scraper.py
```diff
@@ -572,71 +578,69 @@ def get_slot():
...
-                    s3_key,
-                    tag="encoding_version",
-                    value=ENCODER_VERSION,
-                    bucket_name=self.bucket_name,
-                ):
-                    return False
-            os.makedirs(os.path.dirname(local_video_path), exist_ok=True)
-            self.s3_storage.download_file(s3_key, local_video_path)
+            self.s3_storage.download_file(key, video_path)
+            logger.info(f"cache ({key}) downloaded {video_path}")
```
rgaudin, line 594: prefers to keep the log and return after the try block. makes sure we don't capture an error on the logging.

youtube2zim/scraper.py
```diff
@@ -572,71 +578,69 @@ def get_slot():
...
+        # Lets make copy of options dictionary so that it is not shared across concurrent events
+        options_copy = options.copy()
         for video_id in videos_ids:
-            video_location = options["y2z_videos_dir"].joinpath(video_id)
+            video_location = options_copy["y2z_videos_dir"].joinpath(video_id)
             # Reset the skip_download argument
-            options["skip_download"] = False
-            cache_present = None
-            if self.s3_url_with_credentials:
+            options_copy["skip_download"] = False
```
rgaudin, line 621: Now that we have a copy, could we have a copy per video? It would have no memory impact since it's in a loop but it's cleaner and we only set `skip_download` when we need it.

requirements.txt
```diff
@@ -5,4 +5,5 @@ python-dateutil==2.8.0
 jinja2==2.10.1
 babel==2.7.0
 zimscraperlib>=1.0.1,<1.1
-kiwixstorage==0.1
+requests>=2.23.0
```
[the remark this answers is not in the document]
nabinkhadka: I think the latest release is 2.23.0. How can we be sure to have <3.0? How about just `==2.23.0`?

youtube2zim/scraper.py
```diff
@@ -17,12 +17,13 @@
 from pathlib import Path
 import concurrent.futures
 from gettext import gettext as _
+from urllib.parse import urlparse
```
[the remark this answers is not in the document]
nabinkhadka: Didn't know `self.s3_storage.url.netloc`. Removed.

youtube2zim/scraper.py
nabinkhadka: Ok. Did for another occasion.

requirements.txt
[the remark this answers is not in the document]
rgaudin: just use the string is gave you. It ensures it will be `>=2.23` (latest version now) and `<3.0` (next major release, hence API breaking). No `==2.23.0` won't work as other dependencies (storage for instance) might not use just that very same version.

youtube2zim/scraper.py
nabinkhadka: Ok. That will remove another line. Done.

rgaudin:
Found a bug. in `post_process_video`

``` py
# don't reencode if not requesting low-quality and received wanted format
    if not low_quality and src_path.suffix[1:] == video_format:
        return

    dst_path = src_path.parent.joinpath(f"video.{video_format}")

    if recompress:
        recompress_video(src_path, dst_path, video_format, low_quality)
```

 your if statement is unreachable if you don't use low-quality (and you most likely received proper format)

nabinkhadka:
@rgaudin I do not understand how my change introduced the bug. My change was only `if recompress`, it was returning anyways even without that. But I feel like you are saying `if not low_quality and src_path.suffix[1:] == video_format and not recompress:`

rgaudin:
> @rgaudin I do not understand how my change introduced the bug. My change was only `if recompress`, it was returning anyways even without that. But I feel like you are saying `if not low_quality and src_path.suffix[1:] == video_format and not recompress:`

Ah… let me figure this out.

nabinkhadka:
Screen Shot 2020-04-17 at 18 48 19
This is what I purpose if I understand the problem correctly.

rgaudin:
Well it's a bug that was there before. We used to recompress slightly in HQ videos but I do remember writing this shortcut (don't reencode if not LQ). I need to confirm wanted behavior with @kelson42. Nevermind my comment, I'll fix that later. Are you done with everything else?

nabinkhadka:
@rgaudin Yes, but do you want me to send the change as shown in the screenshot attached above? Looks pretty safe to me. Please let me know.

rgaudin:
No, don't.

requirements.txt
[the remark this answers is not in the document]
rgaudin: this has not been fixed

nabinkhadka:
Was planning to send with other commit. But done now.

youtube2zim/entrypoint.py
```diff
@@ -172,6 +172,20 @@ def main():
         help="Custom filter to download videos uploaded on or after specified date. Format: YYYYMMDD or (now|today)[+-][0-9](day|week|month|year)(s)?",
     )
 
+    parser.add_argument(
+        "--optimization-cache",
+        help="URL with credentials to S3 for using as optimization cache",
+        dest="s3_url_with_credentials",
+        default=False,
```
rgaudin, line 179: we should have no default (will default to None)

youtube2zim/entrypoint.py
```diff
@@ -172,6 +172,20 @@ def main():
...
+    )
+
+    parser.add_argument(
+        "--use-any-optimized-version",
+        help="Do not compare the in-code encoder-version with the version stored in the optimization cache",
```
rgaudin, line 184: "Use the cached files if present, whatever the version"

youtube2zim/scraper.py
```diff
@@ -243,10 +260,17 @@ def run(self):
             shutil.rmtree(self.build_dir)
         self.make_build_folder()
 
-        logger.info("testing Youtbe credentials")
+        logger.info("testing Youtube credentials")
         if not credentials_ok():
             raise ValueError("Unable to connect to Youtube API v3. check `API_KEY`.")
 
+        if self.s3_url_with_credentials:
```
rgaudin, line 267: can we have a `logger.info("testing S3 Optimization Cache credentials")` above?

youtube2zim/scraper.py
```diff
@@ -285,8 +309,13 @@ def run(self):
             f"downloading all videos, subtitles and thumbnails (concurrency={self.max_concurrency})"
         )
         logger.info(f"  format: {self.video_format}")
-        logger.info(f"  recompress: {self.low_quality}")
+        logger.info(f"  quality: {self.video_quality}")
         logger.info(f"  generated-subtitles: {self.all_subtitles}")
+        if self.s3_storage:
+            netloc = self.s3_storage.url.netloc
```
rgaudin, line 315: don't use a variable for that, just use it in f-string

youtube2zim/scraper.py
```diff
@@ -550,24 +579,68 @@ def get_slot():
...
+        return True
+
+    def download_video_files_batch(self, options, videos_ids):
         succeeded = []
         failed = []
-        with youtube_dl.YoutubeDL(options) as ydl:
-            for video_id in videos_ids:
-                try:
+        for video_id in videos_ids:
+            # Lets make copy of options dictionary so that changes made in it
```
rgaudin, line 617: think we can remove the comment now that we have a copy per video

youtube2zim/scraper.py
```diff
@@ -550,24 +579,68 @@ def get_slot():
...
-                    )
-                    succeeded.append(video_id)
-                except (youtube_dl.utils.DownloadError, FileNotFoundError):
-                    failed.append(video_id)
+                post_process_video(
+                    video_location,
+                    video_id,
+                    options_copy["y2z_video_format"],
+                    options_copy["y2z_low_quality"],
+                    recompress=not cache_downloaded,  # no need for re-compressing for cached videos
```
rgaudin, line 637: think the comment can be removed

youtube2zim/scraper.py
nabinkhadka: No. We shouldn't because we don't test creds if `if self. s3_url_with_credentials` is not true. So probably below that.

PR #82, opened 2020-04-21
Delete video file if it has no channel associated

satyamtg:
Related to #76 . This ensures that the video files that are downloaded but no longer exist on YouTube are removed.

### DIFF
--- a/youtube2zim/scraper.py
+++ b/youtube2zim/scraper.py
@@ -792,15 +792,12 @@ class Youtube2Zim(object):
             - videos/<videoId>/video.jpg            template
         """
 
-        def remove_unused_videos(used_videos):
-            used_video_ids = []
-            for video in used_videos:
-                used_video_ids.append(video["contentDetails"]["videoId"])
-            videos_dir_contents = [x for x in self.videos_dir.iterdir() if x.is_dir()]
-            for video_dir in videos_dir_contents:
-                if not video_dir.stem in used_video_ids:
-                    logger.debug(f"Removing unused video {video_dir.stem}")
-                    shutil.rmtree(video_dir, ignore_errors=True)
+        def remove_unused_videos(videos):
+            video_ids = [video["contentDetails"]["videoId"] for video in videos]
+            for fname in self.videos_dir.iterdir():
+                if fname.is_dir() and fname.name not in video_ids:
+                    logger.debug(f"Removing unused video {fname.name}")
+                    shutil.rmtree(fname, ignore_errors=True)
 
         def is_present(video):
             """ whether this video has actually been succeffuly downloaded """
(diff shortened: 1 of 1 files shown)


satyamtg:
> I don't really like moving it all to that condition as this will get complicated if we introduce other weird scenarios like this.
> 
> I'd prefer that we leave that loop alone and at the end of that function we clear files out.
> I suggest we loop on videos folder items and compare with what's left in `videos`.
> This way, whatever new filter we add, it will be included.
> 
> FYI, once we move to the pylibzim, this will be rendered useless as we'd be adding videos to the ZIM file on the fly and not creating them on filesystem and then adding it all via zimwriterfs.

@rgaudin have implemented a function to clean unused videos at the end from videos_dir

youtube2zim/scraper.py
```diff
@@ -792,6 +792,16 @@ def make_html_files(self, actual_videos_ids):
             - videos/<videoId>/video.jpg            template
         """
 
+        def remove_unused_videos(used_videos):
```
rgaudin, line 795: could you use `videos` variable name? It would be in its own block and feels more natural as that's what we used.

youtube2zim/scraper.py
```diff
@@ -792,6 +792,16 @@ def make_html_files(self, actual_videos_ids):
...
+            used_video_ids = []
```
rgaudin, line 796: use a comprehension list:
``` py
videos_ids = [v["contentDetails"]["videoId"] for v in videos]
```

youtube2zim/scraper.py
```diff
@@ -792,6 +792,16 @@ def make_html_files(self, actual_videos_ids):
...
+            for video in used_videos:
+                used_video_ids.append(video["contentDetails"]["videoId"])
+            videos_dir_contents = [x for x in self.videos_dir.iterdir() if x.is_dir()]
```
rgaudin, line 799: why create a separate list and then loop on it? If you're not gonna reuse it, just loop directly:
``` py
for fname in self.videos_dir.iterdir():
    if not fname.is_dir():
        continue
    # do your stuff
```

youtube2zim/scraper.py
```diff
@@ -792,6 +792,16 @@ def make_html_files(self, actual_videos_ids):
...
+            for video_dir in videos_dir_contents:
+                if not video_dir.stem in used_video_ids:
```
rgaudin, line 801: use `if xxx not in yyyy` form

rgaudin, line 801: don't use `fname.stem` here as stem is the *filename without extension* part. That won't work if the videoId has a `.` in it (some does).
Instead, use `fname.name` that does what you want.

youtube2zim/scraper.py
satyamtg: Done

rgaudin:
mislead you with my example. renamed `fname` to `path` as it better reflects the actual variable type.

PR #83, opened 2020-04-21
Clearer recompress handling

### LINKED ISSUE #75: should we deal with compression failure?
A [crashcourse run](https://farm.openzim.org/pipeline/5e6ce672ec058ba7715a73e5/debug) failed during the compression of video [k7dTDjRnBqU](https://www.youtube.com/watch?v=k7dTDjRnBqU).

ffmpeg failed to recompress a video and because we're expecting this to succeed, the task failed.

```
Too many packets buffered for output stream 0:1.
[libvorbis @ 0x5632c3166e40] 32 frames left in the queue on closing
Conversion failed!
```
There is a quite old [ffmpeg bug](https://trac.ffmpeg.org/ticket/6375) mentioning this, explaining the reason, why it can't be fixed easily and a workaround for most (but not all) cases: setting `-max_muxing_queue_size` to a larger number than default (`128`).

[One comment](https://trac.ffmpeg.org/ticket/6375#comment:6) mentions getting around that bug by increasing the allocated memory for its GPU. In our context (docker), we don't have GPU but we do set hard limits on RAM.

Given the fact that I couldn't reproduce it on the same file, using the same image, chances are that we're experiencing this due to a RAM issue. RAM will be increased on the recipe and re-run to check.

_reproducing steps_:

``` sh
youtube-dl -o "video.%(ext)s" -f "best[ext=webm]/bestvideo[ext=webm]+bestaudio[ext=webm]/best" k7dTDjRnBqU
docker run -v $(pwd):/data:rw openzim/youtube:2.1.3 ffmpeg -y -i file:/data/video.webm -codec:v libvpx -quality best -cpu-used 0 -b:v 300k -qmin 30 -qmax 42 -maxrate 300k -bufsize 1000k -threads 8 -vf scale='480:trunc(ow/a/2)*2' -codec:a libvorbis -b:a 128k file:/data/video.tmp.webm
```

**Question**: Beside the potential fix for this specific run (resources), should we consider not crashing the scraper on video conversion failures or are we happy with this behavior?

I'm in favor of failing so we get a chance to understand what went wrong and fix the cause but that is dependent on the number of occurrence of conversion failure (it's the first)

rgaudin:
* Fixes #75 by adding `max_muxing_queue_size` to FFmpeg args.
* removes old code for recompressing in high quality (unused)
* better variable names for triggering this
* better wording for S3 logs

### DIFF
--- a/youtube2zim/converter.py
+++ b/youtube2zim/converter.py
@@ -29,7 +29,9 @@ def hook_youtube_dl_ffmpeg(video_format, low_quality, data):
     )
 
 
-def post_process_video(video_dir, video_id, video_format, low_quality, recompress=True):
+def post_process_video(
+    video_dir, video_id, video_format, low_quality, skip_recompress=False
+):
     """ apply custom post-processing to downloaded video
 
         - resize thumbnail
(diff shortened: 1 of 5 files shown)


youtube2zim/converter.py
```diff
@@ -54,17 +56,15 @@ def post_process_video(video_dir, video_id, video_format, low_quality, recompres
     )
 
     # don't reencode if not requesting low-quality and received wanted format
-    if not low_quality and src_path.suffix[1:] == video_format:
+    if skip_recompress or (not low_quality and src_path.suffix[1:] == video_format):
         return
 
     dst_path = src_path.parent.joinpath(f"video.{video_format}")
+    recompress_video(src_path, dst_path, video_format, low_quality)
```
satyamtg, line 63: Need to remove low_quality here. recompress_video() takes in 3 arguments only

PR #92, opened 2020-05-01
Fixes #54 - Hide picture-in-picture toggle

### LINKED ISSUE #54: Picture-in-Picture button is disabled
*@kelson42 commented on Nov 26, 2019, 1:53 AM UTC:*

See:
[image](https://user-images.githubusercontent.com/1029718/69592924-cc785b80-0ff7-11ea-9656-a099bac83564.png)

What is strange is that this control is not displayed at all on the kiwix-serve based version library.kiwix.org/skin-of-color-society_en_all (you have to play the video)

*This issue was moved by [kelson42](https://github.com/kelson42) from [kiwix/kiwix-desktop#310](https://github.com/kiwix/kiwix-desktop/issues/310).*

satyamtg:
Fixes #54 by not displaying PIP toggle if its not supported by the browser. Tested on kiwix-serve versions and Kiwix desktop on Windows and Linux.

@rgaudin could you please confirm if this works on Safari on MacOS?

### DIFF
--- a/youtube2zim/templates/article.html
+++ b/youtube2zim/templates/article.html
@@ -26,7 +26,8 @@
             <video  class="video-js vjs-default-skin"
                     id="video_container"
                     poster="videos/{{ video_id }}/video.jpg"
-                    width="480px" height="270px">
+                    width="480px" height="270px"
+                    data-setup='{"techOrder": ["html5", "ogvjs"], "ogvjs": {"base": "assets/ogvjs"}, "autoplay": {% if autoplay %}true{% else %}false{% endif %}, "preload": true, "controls": true, "controlBar": {"pictureInPictureToggle":false}}'>
                 <source src="videos/{{ video_id }}/video.{{ video_format }}" type="video/{{ video_format }}" />{% if subtitles %}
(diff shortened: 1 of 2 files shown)


kelson42:
@satyamtg nice!

kelson42:
I'm in favour of hidding it eveywhere for the moment because (1) all of this sounds a bit fragile (2) I believe this all should be handled by video.js. But if we do that, I would like to see another ticket to request it's re-introduction if supported by the browser.

rgaudin:
> I'm in favour of hidding it eveywhere for the moment because (1) all of this sounds a bit fragile (2) I believe this all should be handled by video.js. But if we do that, I would like to see another ticket to request it's re-introduction if supported by the browser.

1. absolutely, that would be **very fragile** ; that's why I suggested otherwise.
2. videojs already handles it. It works as expected on native `<video />` for chrome and safari and is hidden in firefox where it will never be supported (firefox's choice).

So there would be no point in opening a new ticket to track this in videojs.

satyamtg:
@rgaudin @kelson42 made the PiP toggle button hidden on the player always. We can now rely on browser feature for Picture in Picture mode.

PR #94, opened 2020-05-18
One zim per playlist

### LINKED ISSUE #93: Support all playlists as individual zims
In [zim-requests #176](https://github.com/openzim/zim-requests/issues/176), we want to create one ZIM file for each of the playlists of a channel.

We might see an increasing need for such requests in the future.

There are at least 2 ways to deal with this:

* _manually_ (or via a script) launch the scraper with `--kind playlist --id xxx`. ie: create one zimfarm recipe for each playlist.
* add a new `--one-zim-per-playlist` option to the scraper.

I suggest we do neither of those but instead add a new, script (different than `youtube2zim`) that would receive the parameters, compute the list of playlists and loop over each to call `youtube2zim`.

This way, we'd keep the scraper simple and zim-oriented while offering this useful feature. The youtube-API part would reuse the code in the `youtube` module obviously.

We may need a new command in Dockerfile so that this is transparent for docker/zimfarm users.

satyamtg:
Fixes #93 . Introduces a new script called youtube2playlistzims which takes in all arguments and runs youtube2zim accordingly to create multiple ZIMs. 

Improvements are of course needed and some formatting changes would be there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     ],
     zip_safe=False,
     include_package_data=True,
-    entry_points={"console_scripts": ["youtube2zim=youtube2zim.__main__:main"]},
+    entry_points={"console_scripts": ["youtube2zim=youtube2zim.__main__:main", "youtube2playlistzims=youtube2zim.extras.youtube2playlistzims.__main__:main"]},
     classifiers=[
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",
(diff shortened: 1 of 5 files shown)


rgaudin:
Thank you for this proposal.

I think it duplicates too much code with youtube2zim and also, it can't serve as a single entrypoint (which the zimfarm needs).

Reusing its concept, I implemented an hopefully simpler solution in #98. 
Let me know what you think.

satyamtg:
Closing this as a better solution is already there in #98

PR #95, opened 2020-05-20
Video compression from zimscraperlib

satyamtg:
This would add video compression using a future version of zimscraperlib when it gets released. Shall remain draft for now.

Also, question - Do we need the reencode_low_quality module? I think it should me moved to some directory such as "extras" which can contain that kind of extra stuff.


rgaudin:
Notice it's WIP but to the answer the question, no we don't need that anymore. It could be moved to `contrib/`

satyamtg:
Moved it. Also updated version in requirements.txt and removed babel from it (installed as a dependency by zimscraperlib)

youtube2zim/scraper.py
```diff
@@ -608,47 +607,97 @@ def download_from_cache(self, key, video_path):
...
+        reencode(
+            src_path, dst_path, enc.to_ffmpeg_args(), delete_src=True, failsafe=False
+        )
+
     def download_video_files_batch(self, options, videos_ids):
         succeeded = []
         failed = []
+
+        # set encoders
+        if self.video_format == "mp4":
```
rgaudin, line 664:
```py
preset = {"mp4": VideoMp4Low}.get(self.video_format, VideoWebmLow)()
```

youtube2zim/scraper.py
```diff
@@ -608,47 +607,97 @@ def download_from_cache(self, key, video_path):
...
-                video_path, key, meta={"encoder_version": ENCODER_VERSION}
+                video_path, key, meta={"encoder_version": enc_version}
             )
         except Exception as exc:
             logger.error(f"{key} failed to upload to cache: {exc}")
             return False
         logger.info(f"uploaded {video_path} to cache at {key}")
         return True
 
+    def post_process_video(self, video_dir, video_id, enc, skip_recompress=False):
```
rgaudin, line 622: use `preset` as that's what it is and `enc` means nothing.
I also think this is a very independent block that doesn't need to be in the scraper class.
We can restore the `low_quality` and `video_format` params and move that off the class.

youtube2zim/scraper.py
```diff
@@ -608,47 +607,97 @@ def download_from_cache(self, key, video_path):
...
+                youtube_dl.utils.DownloadError,
+                FileNotFoundError,
+                subprocess.CalledProcessError,
+            ):
                 failed.append(video_id)
-            if self.s3_storage and not downloaded_from_cache:
-                self.upload_to_cache(s3_key, video_path)
+            else:  # upload to cache only if everything went well
+                if self.s3_storage and not downloaded_from_cache:
+                    self.upload_to_cache(s3_key, video_path, enc.VERSION)
```
rgaudin, line 700: we used to send `"v1"` so sending an int here breaks the S3 storage.
Also as this is a metadata, it must be a string. You may change that in `upload_to_cache`

youtube2zim/scraper.py
```diff
@@ -608,47 +607,97 @@ def download_from_cache(self, key, video_path):
...
         for video_id in videos_ids:
             options_copy = options.copy()
             video_location = options_copy["y2z_videos_dir"].joinpath(video_id)
             downloaded_from_cache = False
             if self.s3_storage:
                 s3_key = f"{self.video_format}/{self.video_quality}/{video_id}"
                 video_path = video_location.joinpath(f"video.{self.video_format}")
-                downloaded_from_cache = self.download_from_cache(s3_key, video_path)
+                downloaded_from_cache = self.download_from_cache(
+                    s3_key, video_path, enc.VERSION
```
rgaudin, line 677: encoder_version is an int here. See comment on line 700.

contrib/reencode_low_quality.py
```diff
@@ -27,22 +30,25 @@ def main(build_path):
         logger.error(f"Build dir `{build_dir}` is not an existing folder.")
         sys.exit(1)
 
-    # import youtube2zim hook
     sys.path = [str(pathlib.Path(__file__).parent.parent.resolve())] + sys.path
```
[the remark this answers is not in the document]
satyamtg: Removed

youtube2zim/scraper.py
```diff
@@ -588,14 +588,14 @@ def get_slot():
...
-    def download_from_cache(self, key, video_path, enc_version):
+    def download_from_cache(self, key, video_path, encoder_version):
         """ whether it successfully downloaded from cache """
         if self.use_any_optimized_version:
             if not self.s3_storage.has_object(key, self.s3_storage.bucket_name):
                 return False
         else:
             if not self.s3_storage.has_object_matching_meta(
-                key, tag="encoder_version", value=enc_version
+                key, tag="encoder_version", value=f"v{str(encoder_version)}"
```
rgaudin, line 598: there is no need for `str` here. f-string does that.

youtube2zim/scraper.py
```diff
@@ -45,6 +44,7 @@
     skip_outofrange_videos,
 )
 from .utils import clean_text, load_json, save_json, get_slug
+from .converter import post_process_video
```
rgaudin, line 47: maybe rename the module to `processing` or something?

PR #102, opened 2020-06-12
Use temporary build_dir

### LINKED ISSUE #101: Use a temporary build dir
For clarity and to allow playlists-mode to create ZIM files on the `--output` folder, we need to stop using a fixed `<output_dir>/build` folder.

This build_dir shall be a real temporary folder (using `tempfile`) and could be specified by an option. The option's help should emphasis that is will receive all the data (requires storage space).

Along with this, playlists mode should be edited to just forward the `--output` option and not specify a custom one.

satyamtg:
This fixes #101 by using a temporary directory as build dir by default, and passing --output directly to youtube2zim in playlists mode. Also, a new option --build-dir is introduced for custom build dir location.

### DIFF
--- a/youtube2zim/playlists/scraper.py
+++ b/youtube2zim/playlists/scraper.py
@@ -141,13 +141,6 @@ class YoutubeHandler(object):
                 # format value using playlists' variables
                 args += [f"--{key}", self.compute_format(playlist, str(value))]
 
-        # ensure we supplied a name
-        if not has_argument("name", args):
-            args += [
-                "--name",
-                self.compute_format(playlist, self.playlists_name),
-            ]
-
         # append regular youtube2zim args
         args += self.extra_args
 
--- a/youtube2zim/playlists/scraper.py
+++ b/youtube2zim/playlists/scraper.py
@@ -15,8 +15,8 @@
 import re
 import sys
 import json
-import shutil
 import pathlib
+import tempfile
 import subprocess
 
 import requests
@@ -24,7 +24,7 @@ from zimscraperlib.logging import nicer_args_join
 
 from ..constants import logger, NAME, YOUTUBE, PLAYLIST
 from ..youtube import extract_playlists_details_from, credentials_ok
-from ..utils import make_build_folder, has_argument
+from ..utils import has_argument
 
 
 class YoutubeHandler(object):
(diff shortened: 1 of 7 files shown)


satyamtg:
Thanks @rgaudin for reviewing the PR so quickly and extensively. Coming to points -


> Understand that when expecting a minor change, seeing many changes, especially touching
> non-related code is misleading

Yup. I'm sorry for that. I shoul've mentioned exact changes and their reasons. My bad. I'll take care of this from now on.

> I think the issue wasn't clear enough that whatever the `--build-dir`, we'd still use a temp subfolder inside to be sure it's really temporary. This means that it can't be reusable and that `--keep` is for debugging only.

Yup. I agree to this. It's a much better thing than what I did. I have seen the changes that you did and they seem amazing. Just I have one question (inline comment)

Coming to your questions - 
> playlists entrypoint to check that variables are used (why?)

Since we are creating multiple ZIMs, I felt that since we were now having ZIMs directly in root of output_dir instead of within subfolders, it becomes that they have unique filenames, that's why I implemented a check of variables so that we ensure unique names. (Should've added check for --zim-file arg too). So, basically that was the idea behind it. I understand that it's not so necessary now.

> --name passing from playlists mode to youtube2zim (why?)

We get the name from the JSON, if not that, we get the self.playlists_name or else None. So, we already have checked the self.playlists_name and I feel setting up again may not make sense (I mean I'm not very sure on this). Moreover, if someone put a JSON, but didn't put the 'name' in it, and he hasn't put the --playlists-name argument, then we might end up failing. So I thought failing early with clear message would be good.

I would love to know your take on this. As for ted, I've done your proposed changes locally and once we have this clear, I'll push them. I'll also comprehensively write what things I have changed and why.

rgaudin:
Thanks, let me try different scenarios on that --name things to be sure.

youtube2zim/playlists/scraper.py
```diff
@@ -145,9 +143,10 @@ def run_playlist_zim(self, playlist):
 
         # ensure we supplied a name
         if not has_argument("name", args):
```
satyamtg: We have already checked the self.playlists_name in line 136. So, why not fail instead of checking self.playlists_name again?

rgaudin: OK you're right. I think we shall remove this check completely but also fix the entrypoint to fail on missing `--playlists-name` (was changed to pass if --metadata-from was provided).
Pushing that now

PR #106, opened 2020-06-19
Add new scraperlib features

satyamtg:
Set this to true explicitly to avoid confusions in the future and removes convert_thumbnail in favour of convert_image() from zimscraperlib

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,10 @@ using zimscraperlib to encode videos
 logging transfer before starting them
 added video_encoding_tester script in contrib/
 --tmp-dir option to set where temp files are downloaded/handled (system temp otherwise)
+thumbnail now converted to JPEG if received as WebP
+thumbnail resize now supports upscaling if original is too small
+thumnail resize and conversion from zimscraperlib
+using zimscraperlib 1.1.1
 
 Version 2.1.5
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ python-slugify==3.0.3
 youtube_dl
 python-dateutil==2.8.0
 jinja2==2.10.1
-zimscraperlib>=1.1.0,<1.2
+zimscraperlib>=1.1.1,<1.2
 requests>=2.23,<3.0
 kiwixstorage>=0.2,<1.0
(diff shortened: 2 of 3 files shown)


rgaudin:
Please explicitly require scrpaerlib over 1.1.1

rgaudin:
Could you replace `convert_thumbnail` now that we have that ability in scraperlib?

satyamtg:
I found out that sometimes youtube-dl returns a webp image with a jpg extension. This is strange. So I added code to convert to JPEG unconditionally

rgaudin:
OK thanks for the head's up. Do you have an URL of such a case that we could add in the comment as this is clearly unexpected.

Also, rewrote variable names to make it clearer (it's a bit tricky!).

Actually, I think your code was wrong as `thumb_src` would be the .webp (if it's a webp) and then you'd be resizing that instead of the .jpg one

satyamtg:
> OK thanks for the head's up. Do you have an URL of such a case that we could add in the comment as this is clearly unexpected.

Used [this](https://www.youtube.com/playlist?list=PL2DQc1tA1YU2sJzyCSjmL8CxG9rN-4wD_) playlist


> Actually, I think your code was wrong as `thumb_src` would be the .webp (if it's a webp) and then you'd be resizing that instead of the .jpg one

Oh. I really missed out on that. Thanks BTW.

rgaudin:
Well there's a single video in that playlist and it's thumbnail as returned by youtube-dl is `https://i.ytimg.com/vi/1ubWja9gP3A/hqdefault.jpg` here. And that's a JPEG.

satyamtg:
Well, this is what actually gets downloaded for the video in the scraper (zipped and attached). The output of libmagic on this file is - 
```'RIFF (little-endian) data, Web/P image, VP8 encoding, 336x188, Scaling: [none]x[none], YUV color, decoders should clamp'
```
Also, if we read it through PIL, the format is WEBP.

[video.zip](https://github.com/openzim/youtube/files/4808247/video.zip)

rgaudin:
Yes, upgraded youtube-dl and do receives WebP in a .jpeg.

PR #110, opened 2020-07-10
Use new zim API from zimscraperlib

rgaudin:
Obviously depends on the merging and release of https://github.com/openzim/python_scraperlib/pull/31

- Use the new `make_zim_file` and get rid of the ZimInfo
- [unrelated] Also move the encoding meta in templates as it's annoying (see commit)


rgaudin:
OK, we'll wait for the download+headers thing to get merged and make a scraperlib release

PR #113, opened 2020-07-17
Fix subtitles display and prevent crash on Hebrew subtitles

### LINKED ISSUE #112: An error occurred: Not a valid iso language name/code
I'm getting this error

and also subtitles embedded in the video do not load when choosing it in the player

satyamtg:
This fixes #112 and has some other subtitle related fixes as follows -

- Fix the namespace for subtitles path in app.js - Subtitles for the main video on the homepage were never loaded, due to incorrect path
- Fix crash on iw (Hebrew) subtitles - Apparently `iw` is not iso-639-1 code. The correct code is `he`. See https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes#HE
- Fix subtitles not showing in article page - Due to incorrect attributes being looked by Jinja2
- Fix subtitle display names and change them to `<native name> - <english name>` - `get_subtitles()` now returns only code and display names as a list of dictionaries.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 2.1.8.dev0
 
-n/a
+fix subtitles not showing on homepage preview video
+fix crash on iw (Hebrew) subtitles
 
 2.1.7
 
--- a/youtube2zim/constants.py
+++ b/youtube2zim/constants.py
@@ -19,6 +19,8 @@ CHANNEL = "channel"
 PLAYLIST = "playlist"
 USER = "user"
 
+YOUTUBELANGMAP = {"iw": "he"}
+
 logger = getLogger(NAME, level=logging.DEBUG)
 
 
--- a/youtube2zim/constants.py
+++ b/youtube2zim/constants.py
@@ -19,7 +19,8 @@ CHANNEL = "channel"
 PLAYLIST = "playlist"
 USER = "user"
 
-YOUTUBELANGMAP = {"iw": "he"}
+# Youtube uses some non-standard language codes
+YOUTUBE_LANG_MAP = {"iw": "he", "es-419": "es"}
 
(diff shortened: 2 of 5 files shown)


rgaudin:
Pushed some modifications:

- renamed the map variable as we never use words concatenated this way.
- also added es-419. Both were taken care of back in the days and were dropped when we switched to scraperlib…
- removed the mapping of the user specified language. We already expressed what we expect in entrypoint. There is no reason for user to specify a youtube-specific code there and should he do that, it will fail early.
- removed the function altogether as this is just getting from a dict
- revamped the subtitle name formatting code so it's clearer and adjusted to match Youtube.com

Let me know if this is OK

satyamtg:
> Pushed some modifications:
> 
> * renamed the map variable as we never use words concatenated this way.
> * also added es-419. Both were taken care of back in the days and were dropped when we switched to scraperlib…
> * removed the mapping of the user specified language. We already expressed what we expect in entrypoint. There is no reason for user to specify a youtube-specific code there and should he do that, it will fail early.
> * removed the function altogether as this is just getting from a dict
> * revamped the subtitle name formatting code so it's clearer and adjusted to match Youtube.com
> 
> Let me know if this is OK



> Pushed some modifications:
> 
> * renamed the map variable as we never use words concatenated this way.
> * also added es-419. Both were taken care of back in the days and were dropped when we switched to scraperlib…
> * removed the mapping of the user specified language. We already expressed what we expect in entrypoint. There is no reason for user to specify a youtube-specific code there and should he do that, it will fail early.
> * removed the function altogether as this is just getting from a dict
> * revamped the subtitle name formatting code so it's clearer and adjusted to match Youtube.com
> 
> Let me know if this is OK

Yup. These actually seem to be better. I wonder why I didn't think of doing it with dict.get() instead of a function.

PR #124, opened 2020-09-25
fixed i18n setup (added pot and mo files)

### LINKED ISSUE #123: Fix i18n setup
The current i18n setup is not correct and has only a po file, whereas there shall be a pot and an mo file as well

rgaudin:
Fixes #123


satyamtg (approved):
:+1:

PR #135, opened 2020-12-30
checking youtubeID in input

### LINKED ISSUE #109: Channel/User/Playlist input checking
https://farm.openzim.org/pipeline/5ef9ad429ec57a810576f6e7/debug failed with an error - 
```
OSError: [Errno 36] File name too long: '/tmp/tmphtp8ytv7/cache/channel_PLr5n3ojAJWjSVnG_EK1xF3rW1Lo0N2qwA,PLr5n3ojAJWjQEiRIuHlRoN7rBDKG6GbvH,PLr5n3ojAJWjRGQ1DnnIqDrIXKuuNydGid,PLr5n3ojAJWjTkqDW49ew1u7vsbVzM5Uub,PLr5n3ojAJWjSVh9mgusLb6npGNnFif_Sw,PLr5n3ojAJWjRSuu4s5Vu1CEN0rkgXZ-VA,PLr5n3ojAJWjRDmRmIVAsD4MSEt7wMSOGr,PLr5n3ojAJWjSNVp6jrlwXPz5MyFArv6oO,PLr5n3ojAJWjRiuPrUAAveWrrqnNPPNzYK.json'
```
This is due to the JSON name being the same as the --id tag it seems.

mahakporwal02:
Fixed #109 
Checking input of youtubeID by user and finishing it early if error occurs.

### DIFF
--- a/youtube2zim/scraper.py
+++ b/youtube2zim/scraper.py
@@ -244,6 +244,9 @@ class Youtube2Zim(object):
         # validate dateafter input
         self.validate_dateafter_input()
 
+        # validate youtube_id input
+        self.validate_ids()
+
         logger.info(
             f"starting youtube scraper for {self.collection_type}#{self.youtube_id}"
         )
@@ -366,6 +369,22 @@ class Youtube2Zim(object):
                 f"YYYYMMDD or (now|today)[+-][0-9](day|week|month|year)(s)."
             )
             raise ValueError(f"Invalid dateafter input: {exc}")
+
+    def validate_ids(self):
+        if self.collection_type == "channel" and (
+            self.youtube_id[:2] != "UC" or len(self.youtube_id) > 24
+        ):
+            raise ValueError("Invalid ChannelId")
+        if self.collection_type == "playlist" and (
+            self.youtube_id[:2] != "PL" or len(self.youtube_id) > 24
+        ):
+            raise ValueError("Invalid PlaylistId")
+        if self.collection_type == "user" and (
+            self.youtube_id[:2] == "PL"
+            or self.youtube_id[:2] == "UC"
+            or len(self.youtube_id) > 24
+        ):
+            raise ValueError("Invalid UserId")
 
     def prepare_build_folder(self):
         """ prepare build folder before we start downloading data """
--- a/youtube2zim/scraper.py
+++ b/youtube2zim/scraper.py
@@ -245,7 +245,7 @@ class Youtube2Zim(object):
         self.validate_dateafter_input()
 
         # validate youtube_id input
-        self.validate_ids()
+        self.validate_id()
 
         logger.info(
             f"starting youtube scraper for {self.collection_type}#{self.youtube_id}"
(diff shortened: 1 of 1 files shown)


rgaudin (changes requested):
Thank you.

The first issue I see is that this code doesn't allow multiple playlists while the current code does… Please adjust the check.

I'd like a comment with a link pointing to those format (`UC`, `PL`).

Are you sure about the length of all those IDs? User and Channel can be customized I believe so this might be different than the generated ones.

mahakporwal02:
@rgaudin I think API request would be more consistent method for implementing this, sorry for wasting your time.

rgaudin:
> @rgaudin I think API request would be more consistent method for implementing this, sorry for wasting your time.
> I'll try to implement it accordingly

I think you're getting off track. We already fail on incorrect IDs specifically when we try to retrieve info from the API. What you are now suggesting won't help much.

What the issue is about is failing very early in case it's an *obvious* type-ID combination error.

rgaudin:
OK Thanks ; as you've seen it's become very minimal. A new quick check we can add is looking if there are commas `,` in the ID. There should not be any for `channel` nor `user` mode. Commas are only for passing a list of playlists in `playlist` mode. 

If there are commas, it might be good to remove spaces from the ID as a common mistake is to separate playlists IDs with `, ` instead of `,`. Also, spaces are not allowed in IDs anyway.

mahakporwal02:
@rgaudin Pls have a look, I have updated the PR

youtube2zim/scraper.py
```diff
@@ -244,6 +244,9 @@ def run(self):
         # validate dateafter input
         self.validate_dateafter_input()
 
+        # validate youtube_id input
```
rgaudin, line 247: un-necessary comment

youtube2zim/scraper.py
```diff
@@ -367,6 +370,15 @@ def validate_dateafter_input(self):
             )
             raise ValueError(f"Invalid dateafter input: {exc}")
 
+    def validate_id(self):
+        self.youtube_id = self.youtube_id.replace(" ", "")
```
rgaudin, line 374: comment here and group it all:

```py
        # space not allowed in youtube-ID
        self.youtube_id = self.youtube_id.replace(" ", "")
        if self.collection_type == "channel" and len(self.youtube_id) > 24:
            raise ValueError("Invalid ChannelId")
        if ";" in self.youtube_id and self.collection_type == "playlist":
            raise ValueError("Invalid YoutubeId")
```

youtube2zim/scraper.py
```diff
@@ -244,6 +244,9 @@ def run(self):
         # validate dateafter input
         self.validate_dateafter_input()
 
+        # validate youtube_id input
+        self.validate_id()
```
rgaudin, line 248: I'd move that before the dateafter one as this is the main user input so makes more sens to test that first

rgaudin (changes requested):
Thanks ; it's almost there. can you make those tiny formatting changes as well?

youtube2zim/scraper.py
mahakporwal02: In the second "if" condition I think there should be a "," instead of ";" and "!=" instead of "==" or pls correct me if I am wrong somewhere

rgaudin:
you're right

mahakporwal02:
> you're right

Thanks

rgaudin (approved):

Issue #171, opened 2023-02-06
Can zimscraperlib 2.0 be used?


holta:
Thanks for [zimscraperlib 2.0](https://github.com/openzim/python-scraperlib/releases/tag/v2.0.0) was released 2 months ago!

Question: Can we begin allowing this as part of youtube2zim?

https://github.com/openzim/youtube/blob/2897b6d39fe3bf6b9409b0cc8ce022c188653ac6/requirements.txt#L8

2022-11-08 background:

- "Using zimscraperlib 1.8.0" https://github.com/openzim/youtube/commit/c9938a65a5683431e952caae645afc37d672f923

rgaudin:
👍

PR #172, opened 2023-02-07
requirements.txt: zimscraperlib>=2.0.0,<2.1.0

holta:
Thanks to @deldesir who quickly tested this with zimscraperlib 2.0.0

@rgaudin & @kelson42 is this what's needed?

- #171

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ python-dateutil==2.8.0
 jinja2>=2.11,<3.0
 # jinja2 dependency (https://github.com/pallets/markupsafe/issues/284)
 MarkupSafe==2.0.1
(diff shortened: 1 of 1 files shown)


rgaudin (changes requested):
The idea is to bump is to the version you want to use yet define a range to allow for its future minor versions to be used.
So it would be:

```
zimscraperlib>=2.0.0,<2.1.0
```

holta:
> The idea is to bump is to the version you want to use yet define a range to allow for its future [PATCH] versions to be used.

I assume you mean PATCH versions rather than MINOR, per https://semver.org/#semantic-versioning-200

Done!

rgaudin (approved):
LGTM

rgaudin:
Yes, sorry 😉

Issue #184, opened 2023-11-16
urllib3 versions are conflicting


benoit74:
In [canadian_prepper recipe](https://farm.openzim.org/recipes/canadian_prepper), we just saw with @RavanJAltaie and @Popolechien that we keep getting an error about ssl ciphers.

See for instance [last task](https://farm.openzim.org/pipeline/c9b89099-7549-4df2-875d-29d6400afaca/debug) where everything else is supposed setup properly.

STDERR:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
botocore 1.29.5 requires urllib3<1.27,>=1.25.4, but you have urllib3 2.1.0 which is incompatible.
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
WARNING: You are using pip version 22.0.4; however, version 23.3.1 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
Traceback (most recent call last):
  File "/usr/local/bin/youtube2zim", line 33, in <module>
    sys.exit(load_entry_point('youtube2zim==2.1.18', 'console_scripts', 'youtube2zim')())
  File "/usr/local/lib/python3.8/site-packages/youtube2zim-2.1.18-py3.8.egg/youtube2zim/__main__.py", line 13, in main
    from youtube2zim.entrypoint import main as entry
  File "/usr/local/lib/python3.8/site-packages/youtube2zim-2.1.18-py3.8.egg/youtube2zim/entrypoint.py", line 10, in <module>
    from .scraper import Youtube2Zim
  File "/usr/local/lib/python3.8/site-packages/youtube2zim-2.1.18-py3.8.egg/youtube2zim/scraper.py", line 28, in <module>
    from kiwixstorage import KiwixStorage
  File "/usr/local/lib/python3.8/site-packages/kiwixstorage-0.8.3-py3.8.egg/kiwixstorage/__init__.py", line 27, in <module>
    import boto3
  File "/usr/local/lib/python3.8/site-packages/boto3-1.26.5-py3.8.egg/boto3/__init__.py", line 17, in <module>
    from boto3.session import Session
  File "/usr/local/lib/python3.8/site-packages/boto3-1.26.5-py3.8.egg/boto3/session.py", line 17, in <module>
    import botocore.session
[...]

PR #185, opened 2023-11-16
Unset `metadata_from` + rollback to Python 3.10

benoit74:
# Rationale
- we should release a 2.2.0 asap with current `main` status since this would solve #184 (which is blocking all scrapers) but:
  - there is a strange error about `metadata_from` when running the scraper with `youtube2zim-playlists` ; it is obviously harmless to simply disable this code since it is not possible to setup the `metadata_from` which current code anyway
  - `main` is in fact not correct, it is using Python 3.11 while we are still at `zimscraperlib` 2.0 which supports only Python 3.10 (don't know why this is not causing a problem in Docker build + why it does not raised a red flag before)

# Changes
- set `metadata_from` to always be `None` for now
- rollback to Python 3.10
- align `zimscraperlib` version constraints in hatch dependencies with project version (just in case)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11-bookworm
+FROM python:3.10-bookworm
 LABEL org.opencontainers.image.source https://github.com/openzim/youtube
 
 # Install necessary packages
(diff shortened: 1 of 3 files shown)


benoit74:
I propose to ignore codecov warning, since the file is not covered at all anyway.

benoit74:
I just fixed the PR title which was wrong indeed.

rgaudin (approved):
LGTM

Issue #189, opened 2024-01-01
Update the ReadMe.md file


jo780-full:
The installation process is missing a crucial step, causing issues with the get_js_deps.sh file not being downloaded during the installation of youtube2zim using pip3. To address this, I suggest updating the installation section in the readme with the following steps:
**FIX**
1. clone the repository
`git clone https://github.com/openzim/youtube`
`cd youtube`

2. Run the get_js_deps.sh script:
`./get_js_deps.sh`

3.Copy assets to the virtual environment:
`cp -Rf /path_to_your_directory/youtube/src/youtube2zim/templates/assets /path_to_your_env/lib/python3.10/site-packages/youtube2zim/templates/
`

This should fix the no videojs-ogvjs.js  error .

jo780-full:
Hey @benoit74 ,
I don't think the `get_js_deps.sh` is being run upon installation the assets aren't present in the virtual environment when i run pip install youtube2zim in my virtual environment (.venv) all the required assets aren't present in `.venv/lib/python3.10/site-packages/youtube2zim/templates/assets/ ` . That's what is causing the issue.

jo780-full:
how can we get it to run during installation ?

a simple workaround is to download the script
build the dependencies by running the script and then copying them into the package directory within the virtual environment

but i don't think it's very user-friendly as it involves a lot of manual steps to set up.

benoit74:
Hi, thank you for reporting that.

Now that I'm in front of my laptop and the whole codebase, I realize that the problem is in fact "simply" that we did not updated the documentation after the migration to `hatch` instead of `setuptools`.

Please read the new README.md and tell me if it is still not clear. 

Sorry for the time you lost adjusting the README for this lack of update.

PR #190, opened 2024-01-01
update: added a simple workaround for setup issues

jo780-full:
This  is a simple work around for  #189.
Following the installation process  in the current `README.md` ,all the required assets aren't built.  For instance ,there is an  error that states  `video-ovjs.js` file is not availble in the tmp/templates/assets folder I have had to build them separately and copy them into the library file within my virtual environment. 

This PR contains the required steps for the same.But this is not an ideal fix, considering the user uses pip3 to install the library packages. This is only an interim change; the fix should be to build the dependencies while installing youtube2zim from pip

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,27 @@ Here comes a few different ways to install `youtube2zim`.
 [virtualenv](https://virtualenv.pypa.io) to avoid installing software
 dependences on your system.
 
+
+
+
 ```bash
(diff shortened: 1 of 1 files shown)


benoit74:
This has been fixed by https://github.com/openzim/youtube/commit/26f5c712a49ebe7cbffe2cc2e8ad597ff9ea99dc indeed

PR #191, opened 2024-01-10
update: added a simple workaround for setup issues

### LINKED ISSUE #175: Upgrade python-scraperlib to 3.x, including CLI support for description / long_description flags
See https://github.com/openzim/overview/issues/21

jo780-full:
FIX #175 
Hi @benoit74 , This is the first version of the changes , I still need to add the validation of metadata check before the scraper starts the downloading process. However, I have made a few changes that makes the scraper compatible with the new version of the zimscraper library.I have some doubts regarding the validation process, I have messaged regarding the same.


benoit74:
Hi, I will review the PR only once the code is ready to be merged. Here your PR does not change anything. I moved the PR to Draft for now.

If you have concerns that need to be discussed, please ask questions here (or in the issue) so that they can be discussed in open.

jo780-full:
@benoit74 , i added checks for ASAP in the scraper.py by calling  `validate_metadata()`. Kindly do check and let me know,sorry it took a while because I had some other hold up :)

benoit74 (changes requested):
Thank you, unfortunately this is still far from being OK:
- the description you validate is not the one which is finally used for creating the ZIM, the description is further updated in `update_metadata` which is called way after your validating it
- you do not validate the long_description
- the new function `populate_metadata` is useless (it just computes a dictionary, there is no reason to create a new function for it) and misleading (I first understood that you were populating metadata in the ZIM or in the scraper properties while you just computes something
- your PR is "bad" (sorry):
  - PR title does not express at all what this PR is intended to do
  - your branch contains irrelevant commits ; you should have only few relevant commits on top of `main`, and no merge commits at all (i.e. get rid of d5e5e02948ab07f0e4c4e712eff5af5d99d4bda9 and 7ecd03015713d5a6992e7b20f478b8058f933996, you will have to force push but this is not a problem here since the current situation is worse than a force push, you have to clean this up)
- your code computes description and long_description from auto_description, and then uses `compute_description`, which is a non-sense, `compute_description` is sufficient

This should be enough to compute description and long description:
```
      auto_description = (
            clean_text(self.playlists[0].description)
            if self.is_playlist and len(self.playlists) == 1
            else clean_text(main_channel_json["snippet"]["description"])
        ) or "-"
        self.title = self.title or auto_title or "-"
        self.description, self.long_description = compute_descriptions(
            default_description=auto_description,
            user_description=self.description,
            user_long_description=self.long_description,
        )
```

But this should be done:
- before validating metadata
- as soon as we have sufficient data, and in any case before `download_video_files` which takes most of the execution time ; you should probably move `update_metadata` sooner in the execution path, you have to check what is needed in `update_metadata` and when this information is retrieved, and move your validation code right after `update_metadata` (and you should probably move all this validation logic in a dedicated `validate_metadata` function

jo780-full:
Sure, Ill try working on it. Thanks for the feedback

benoit74:
Closing this, it has been staled for weeks and similar work is now mostly ready to be merged in #202 (and I'm sure it will be merged in the coming days)

PR #196, opened 2024-03-13
Fixes #192: Update default publisher to 'openZIM'

### LINKED ISSUE #192: Harmonize default publisher to `openZIM`
As required in https://github.com/openzim/zim-requests/issues/779, the default ZIM publisher must be `openZIM` (mind the case)

nida-imran173:
### Description
This pull request resolves the issue by updating the default publisher from "Kiwix" to "openZIM".

### Changes Made

- Updated the default publisher value to "openZIM".

Fixes https://github.com/openzim/youtube/issues/192

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Changed default publisher metadata to 'openZIM'
+
 ## [2.2.0] - 2023-11-17
 
 ### Changed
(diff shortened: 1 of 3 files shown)


benoit74 (changes requested):
Almost good, thank you. Just one more thing I forgot in previous review (because I also tend to forget to do it myself, but let's try to give you good habits): we need to update the Changelog. Add an entry in the `## [Unreleased]` section (not in a version section, we will assign a version to this change during the release process). See https://github.com/openzim/ted/pull/165/files for a good example.

benoit74 (approved):

PR #207, opened 2024-05-27
Move scraper to subfolder and initialize new Vue.js project

### LINKED ISSUE #205: Move scraper to subfolder and initialize new Vue.js project
As part of #165, we need to move the scraper related files into a subfolder called `scraper` and initialize a new Vue.js project in a folder called `zimui`.

This will be as the same approach followed in the [Kolibri UI revamp](https://github.com/openzim/kolibri/tree/new_ui).

dan-niles:
Changes:
- Scraper (Python code) has been moved to the `scraper` subfolder
- Vue.JS is now used as main UI framework; all of its code is in the `zimui` subfolder; it is rendered with Vite to produce a static website.
- Vuetify has been installed and setup
- QA and Tests workflows have been adapted
  - to the new folder structure
  - to also QA and Test the Vue.JS part
- precommit hooks have been configured for the Vue.JS part
 - Dockerfile has been adapted to first build the Vue.JS part in a dedicated stage and then embed the generated files into the final Python-based image

Closes #205


scraper/src/youtube2zim/entrypoint.py
```diff
@@ -86,6 +87,15 @@ def main():
         default=False,
     )
 
+    parser.add_argument(
+        "--zimui-dist",
+        type=str,
+        help=(
+            "Directory containing Vite build output from the Zim UI Vue.JS application"
+        ),
+        default=os.getenv("KOLIBRI_ZIMUI_DIST", "../zimui/dist"),
```
benoit74, line 96: `KOLIBRI_ZIMUI_DIST` is probably not the right environment variable name ;)

benoit74 (changes requested):
Few comments inline.


Some other points:
- you must mention in your issue that:
  -  you've started the move to "on-the-fly" items addition to the ZIM (I've just opened https://github.com/openzim/youtube/issues/209 which is about it, the move is not yet finished since you have to fix all commented code)
  - from this point, the scraper is intentionally quite broken in the sense that it does not create a usable ZIM anymore since we had to comment out significant code portions to move to the new ZIM ui and creator
- add a disclaimer at the beginning of README.md explaining that for now the scraper is undergoing significant rework for 3.0 release and the `main` branch is not very stable ; `v2` branch is available (I just created it) should an important patch be pushed before v3 release
- do we really need to keep the `zimui/README.md`, it does not provide much value from my PoV ?

We should also open new issues to not forget that we have to (no need to do it in current PR, can be tackled afterwards):
- add a `CONTRIBUTING.md` at the root of the project explaining how to develop the `zimui`, just like we have at https://github.com/openzim/kolibri/blob/new_ui/CONTRIBUTING.md
- fix code coverage upload for Python part ; looking at the CI it looks like it fails to find the coverage report, it is probably broken as well in kolibri ; since we have almost zero test anyway, it is not an urgent problem to solve ^^

Dockerfile
```diff
@@ -14,24 +21,30 @@ RUN apt-get update \
...
 # Install + cleanup
-RUN pip install --no-cache-dir /src \
- && rm -rf /src
+RUN pip install --no-cache-dir /src/scraper \
+ && rm -rf /src/scraper
+
+# Copy zimui build output
+COPY --from=zimui /src/dist /src/zimui
+
+ENV KOLIBRI_ZIMUI_DIST=/src/zimui
```
benoit74, line 48: Same remark ^^

scraper/src/youtube2zim/entrypoint.py
dan-niles: Oops 😅 Fixed it in [304e7c3](https://github.com/openzim/youtube/pull/207/commits/304e7c367e9ff184ea5865ae1499f880115766d4)

Dockerfile
dan-niles: Fixed it in [304e7c3](https://github.com/openzim/youtube/pull/207/commits/304e7c367e9ff184ea5865ae1499f880115766d4)

PR #232, opened 2024-06-18
Finish removal of locale related code

### LINKED ISSUE #226: New UI: Finish removal of locale for now
In #212 we removed the "locale" thing since it was anyway broken.

We however forgot to remove the call the `setlocale` in `scraper.py`, which breaks the scraper since locale isn't available.

This should be cleaned-up as well until we decide how to implement translations in new UI.

dan-niles:
This PR removes the `--locale` argument from entrypoint.py and some code from scraper.py using the `setlocale` method.

- Closes #226


benoit74 (approved):
LGTM, thank you.

Issue #235, opened 2024-06-21
 Migrate to scraperlib 3.4.0 and VP9 presets


benoit74:

benoit74:
Done for v2, still to do for v3

PR #236, opened 2024-06-21
Upgrade to scraperlib v3.4.0 to use VP9

benoit74:
Partial implementation of #235 

- partial because this PR is an update on the `v2` branch since v3 is not yet ready
- this also needs to be backported to `main` branch as well
- tests done
  - run again the scraper => all videos are reencoded with VP9 and reuploaded to cache
  - run again the scraper a second time => all videos are downloaded from cache and are encoded with VP9
- also implement bootstrap v1.0.2 (fixes to code coverage upload)


rgaudin (approved):
LGTM

Issue #233, opened 2024-06-19
Set autoplay option to always be true for videos in new UI


dan-niles:
With the implementation of #229, the video.js player has been configured to have autoplay set to false by default. The entrypoint.py script includes an argument called `--autoplay`, which allows the user to enable or disable autoplay functionality.

We need to update the scraper so that it can convey this autoplay configuration to the Vue.js UI.

Kaif9999:
Hey I wanted to contribute to this issue is it still open.

kelson42:
@Kaif9999 Hi, it's open but assigned to @dan-niles, maybe you can find one which is not assigned?

Kaif9999:
Hi, also wanted to ask are there any issues I can work with I am a beginner to open source with the knowledge of Python have built a few small projects in it, I am also proficient in html, css, javascript

benoit74:
@Kaif9999 youtube scraper is indeed undergoing a significant effort of rewriting, so this issue is assigned to Dan because he is the best one to tackle this issue.

Please also prefer to consider issues tagged "good first issue" and not assigned to someone (or without activity by this person for months) preferably. Preferably because it is ok to ask for contributing on issues not matching these criteria, but the chances we say "no" are higher obviously ^^

More information on how to contribute at www.kiwix.org/code

dan-niles:
In #237 we decided to remove the `--autoplay` argument and set `autoplay: true` always.

PR #237, opened 2024-06-21
Add playlist panel to video player page

### LINKED ISSUE #216: UI Revamp: Playlist Video Player Page
This is a subtask of #165 

The "Playlist Video Player" page is where a video from a playlist can be viewed along with other videos in the playlist, by the user. The UI redesign for this page will be based on the following designs:

- Desktop View:

- Mobile View:

dan-niles:
This PR adds a playlist panel to the video player page.
- Added a playlist panel that displays the videos in a playlist (Playlist is loaded through the `list` query parameter in URL)
- Added player control buttons to loop (Loop video/playlist) and shuffle playlist.
- Updated error handling for unknown videos, playlists and paths.

Screenshot:
image

Closes #216

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Create video player page in new Vue.js UI (#215)
 - Add support for variable playback speed in video player (#174)
 - Updgrade to zimscraperlib 3.4.0 (including **new webm encoder presets to migrate to VP9 instead of VP8**) (#204)
+- Add playlist panel for playing videos in a playlist (#216)
 
 ## [2.3.0] - 2024-05-22
 
--- a/zimui/cypress/e2e/video_player_page.cy.ts
+++ b/zimui/cypress/e2e/video_player_page.cy.ts
@@ -22,4 +22,32 @@ describe('video player page', () => {
     cy.contains('.video-channel', 'openZIM_testing')
     cy.contains('.video-description', 'This is a short video of a timelapse.')
   })
+
+  it('loads the playlist panel', () => {
+    cy.intercept('GET', '/playlists.json', { fixture: 'channel/playlists.json' }).as('getPlaylists')
+    cy.intercept('GET', '/playlists/timelapses-QgGI.json', {
+      fixture: 'channel/playlists/timelapses-QgGI.json'
+    }).as('getPlaylist')
+    cy.intercept('GET', '/videos/timelapse-9Tgo.json', {
+      fixture: 'channel/videos/timelapse-9Tgo.json'
+    }).as('getVideo1')
+    cy.intercept('GET', '/videos/cloudy_sky_time_lapse-k02q.json', {
+      fixture: 'channel/videos/cloudy_sky_time_lapse-k02q.json'
+    }).as('getVideo2')
+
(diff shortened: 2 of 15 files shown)


benoit74 (commented):
Looks good, works well, thank you!

Only a small interrogation. I'm not familiar at all with the `nextTick` API, and I fail to find examples on the net on how to use this properly. Do we really need the two calls to `nextTick`, once in the `watch` and once in the method itself?

benoit74:
Maybe one question for @Popolechien: we have a setting at ZIM level regarding autoplay, which is used to decide if we want to autoplay the video every time a page is opened. By default, it is off.

However, with the new UI in playlist mode we will now automatically move from one video to the next when the video is over. Do we still want to take into account the autoplay setting? I.e. if autoplay setting is false, when a video finishes, the UI moves automatically to the next video but video does not start. Or do we want to always start the next video automatically in such a situation?

Popolechien:
What does Youtube do? Unless we can make a compelling argument that our users are different from theirs, we should mimic YT's behaviour as much as possible (which may, at times, include things we at a personal level do not particularly enjoy).

Autoplay it is, then.

benoit74:
Agreed.

Does it means we do not really need the "autoplay" setting at scraper level and could even imagine to get rid of it? Do you recall when / why it was introduced? I imagine it could be to save bandwidth (avoid loading the video if the user finally decides he does not want to see it) but I'm not sure it makes lot of sense in terms of usability.

Popolechien:
> Do you recall when / why it was introduced?

I think you wildly overestimate my technical involvement or even my ability to understand why certain technical choices are made. But thank you for that :-) 

I am pretty sure I don't understand the difference between autoplay and autoplay at scraper level, so I'll focus on the end result and simply ask that things, well, _autoplay_.

dan-niles:
> Only a small interrogation. I'm not familiar at all with the `nextTick` API, and I fail to find examples on the net on how to use this properly. Do we really need the two calls to `nextTick`, once in the `watch` and once in the method itself?

@benoit74 Oops that's a mistake. Thanks for pointing it out. I must have duplicated it when doing some debugging and forgot to remove it. Removed it in [59b89cd](https://github.com/openzim/youtube/pull/237/commits/59b89cd3f3fa30754d98a5ce0dab11f40faa2802).

[nextTick()](https://vuejs.org/api/general.html#nexttick) is used so that the scrolling in the playlist panel happens after the DOM updates for loading the video are completed.

dan-niles:
If there is anything else to fix please let me know or else I can squash and rebase. Should I remove the `--autoplay` argument and make autoplay always true? or do we create a seperate issue for this?

benoit74:
> If there is anything else to fix please let me know or else I can squash and rebase. 

Please squash and rebase

> Should I remove the --autoplay argument and make autoplay always true? or do we create a seperate issue for this?

Please update https://github.com/openzim/youtube/issues/233 by changing title and add a new comment saying that we've finally decided to `autoplay: true` and remove CLI argument, referring to the discussion in this PR. And do the change in other PR for clarity.

benoit74 (approved):
LGTM, please squash and rebase

dan-niles:
Squash and rebase completed

benoit74 (changes requested):
Looks like rebase is not finished, your branch is still not on top of `main` latest commit

benoit74 (approved):
LGTM, thank you

PR #244, opened 2024-06-25
Remove `autoplay` CLI argument

### LINKED ISSUE #233: Set autoplay option to always be true for videos in new UI

dan-niles:
- Remove `--autoplay` CLI argument and set videos to always autoplay.
- Remove some unused files from the `templates/assets` folder.

Closes #233


benoit74 (changes requested):
Please add a CHANGELOG entry, otherwise LGTM. You can directly amend your commit, it is easy to read ^^

benoit74 (approved):
LGTM

PR #250, opened 2024-06-28
Fix title in index.html to be ZIM title

### LINKED ISSUE #227: New UI: HTML page title should be the ZIM title
The HTML page title is currently a constant "YouTube in a ZIM". This is displayed for instance in browser tabs. 

This title should be "dynamically"  aligned with the ZIM title (passed to the scraper CLI or computed dynamically).

The way the title is set is important, because ideally it should be available in the HTML when page loads because most (all?) readers use this HTML page title to render it somewhere in the UI but most (all?) do not respond to dynamic modification of the page title via JS code. But maybe this is just too complex and we should rather open issues in readers to support dynamic updates of the page titles. TbC.

dan-niles:
Fix #227 

Replace the `<title>` tag (using RegEx) in index.html of new Vue.js UI to be the ZIM title.

### DIFF
--- a/scraper/src/youtube2zim/scraper.py
+++ b/scraper/src/youtube2zim/scraper.py
@@ -428,11 +428,27 @@ class Youtube2Zim:
                 continue
             path = str(Path(file).relative_to(self.zimui_dist))
             logger.debug(f"Adding {path} to ZIM")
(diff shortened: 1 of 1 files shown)


benoit74 (approved):
LGTM

PR #251, opened 2024-06-28
Display playlist panel for videos played from homepage

### LINKED ISSUE #241: Show playlist panel instead of arrows for videos from the `Videos` tab
@Popolechien In our initial discussion we decided to add arrows to videos (#215) played from the `Videos` page (#213) so that users can navigate to different videos once they have finished viewing the one they clicked on. Something like this:

image

With #237 a playlist panel listing all the videos in a playlist is being added. In this, users can view the thumbnails and title of all videos in a playlist. I propose we use this same playlist panel in the single video page so that users explore more videos in the ZIM easily. (With this they have the option to also loop/shuffle videos).

image

This is also much easier to implement compared to adding arrows buttons to the side of videos, since we have the code ready in #237. What do you think? Shall we do it in this method?

dan-niles:
Close #241

Modified `VideoGrid` and `VideoCard` components to accept a `playlistSlug` prop, so that the playlist panel can be displayed whenever videos from the homepage are played.

### DIFF
--- a/zimui/src/components/channel/tabs/VideosTab.vue
+++ b/zimui/src/components/channel/tabs/VideosTab.vue
@@ -48,5 +48,5 @@ onMounted(() => {
     :count-text="playlist?.videos.length === 1 ? 'video' : 'videos'"
     icon="mdi-video-outline"
   />
-  <video-grid v-if="videos" :videos="videos" />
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM, I will take care of rebase + merge

PR #255, opened 2024-07-01
Fix side margins on tablet screens in zimui

### LINKED ISSUE #252: New UI: Remove side margins in UI for tablet screens
Currently on tablet screens there are margins on the left and right side of the UI. These are good to have on large monitors so that the UI is always centered on the screen, but for medium sized screens like tablet screens, this is a waste of space. 

I propose we remove these margins so that more content is visible on tablet screens.

With margins on tablet screen (Current UI):
image
image

Without margins on tablet screen (Proposed UI):
image
image

dan-niles:
Changes:
- Fix side margins on tablet screens so that the content fills the entire screen.
- Modify playlist panel height calculation so that the panel fits screen vertically.

Before:
image

After:
image

Fix #252

### DIFF
--- a/zimui/src/components/channel/ChannelHeader.vue
+++ b/zimui/src/components/channel/ChannelHeader.vue
@@ -1,11 +1,14 @@
 <script setup lang="ts">
 import { onMounted, ref } from 'vue'
+import { useDisplay } from 'vuetify'
 
 import AboutDialogButton from '@/components/channel/AboutDialogButton.vue'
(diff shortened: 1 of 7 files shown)


benoit74 (approved):
Not tested in real conditions, but it looks coherent, so LGTM, I assume you tested it significantly at various resolutions

We will need anyway a throughout review of look and feel at various screen sizes once new UI is ready.

Issue #253, opened 2024-06-28
Write integration tests for the scraper


dan-niles:
It would be beneficial if we could write integration tests for the YouTube scraper similar to the one in [zimit](https://github.com/openzim/zimit/blob/main/tests-integration/integration.py).

benoit74:
> Do you have any suggestions for any other integration tests that might be useful?

I would suggest to test for subtitles (there is only one in openZIM test channel, but still good to check it is still there). 

Nothing else to suggest, I think this will be a good start and it is always easier to add tests to an existing suite when we discover a bug rather than having to start from scratch. From my PoV job's is sufficient since this goal is reached.

PR #258, opened 2024-07-01
Setup integration tests and workflow

dan-niles:
This PR adds integration tests for the scraper and sets up the necessary GitHub CI workflow.

#### Related issue:
- #253

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -86,15 +86,21 @@ jobs:
           cd zimui
           $(yarn bin)/cypress run
 
-  build-docker:
+  # this job replaces the standard "build_docker" job since it builds the docker image
+  run-integration-tests:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
 
       - name: Ensure we can build the Docker image
         run: |
-          docker build -t testimage .
+          docker build -t youtube2zim .
 
-      - name: Ensure we can start the Docker image
-        run: |
-          docker run --rm testimage
+      - name: Run scraper
+        env:
+          YOUTUBE_API_KEY: ${{ secrets.YOUTUBE_API_KEY }}
+          OPTIMIZATION_CACHE_URL: ${{ secrets.OPTIMIZATION_CACHE_URL }}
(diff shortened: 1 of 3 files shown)


dan-niles:
@benoit74 Can you create the secret for `YOUTUBE_API_KEY`?

.github/workflows/Tests.yml
```diff
@@ -86,15 +86,24 @@ jobs:
...
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
 
       - name: Ensure we can build the Docker image
         run: |
-          docker build -t testimage .
+          docker build -t youtube2zim .
 
       - name: Ensure we can start the Docker image
```
benoit74, line 99: Do you see any significant advantage in keeping this step isolated? This was meant to be a minimal integration test (can we at least start the container and the process within) but now that we have more thorough integration tests, I'm not sure there is any benefit to keeping this step

.github/workflows/Tests.yml
```diff
@@ -86,15 +86,24 @@ jobs:
...
         run: |
-          docker run --rm testimage
+          docker run --rm youtube2zim
+
+      - name: Run scraper
+        env:
+          YOUTUBE_API_KEY: ${{ secrets.YOUTUBE_API_KEY }}
+        run: docker run -v $PWD/output:/output youtube2zim youtube2zim --api-key "$YOUTUBE_API_KEY" --type channel --id "UC8elThf5TGMpQfQc_VE917Q" --name "openZIM_testing" --zim-file "openZIM_testing.zim"
```
benoit74, line 106: I think we should also pass an optimization cache URL, otherwise the CI will spend too much time reencoding videos. WDYT? If you agree, I will create a dedicated Wasabi user for this, and we will need a new CI secret

benoit74 (changes requested):
See comments, small interrogations

.github/workflows/Tests.yml
dan-niles: Ah I see. It is redundant. I'll remove it.

.github/workflows/Tests.yml
dan-niles: Sure sounds like a good idea. I'll update the workflow command.

.github/workflows/Tests.yml
dan-niles: Removed in [bb30caf](https://github.com/openzim/youtube/pull/258/commits/bb30caf8be6113c27ff74c80aa2f2ec99c5cdde9)

benoit74 (approved):
LGTM

PR #259, opened 2024-07-02
Add support for ogv.js in video player

### LINKED ISSUE #218: Install all JavaScript dependencies through Yarn
Right now all the JavaScript dependencies in [openzim.toml](https://github.com/openzim/youtube/blob/main/scraper/openzim.toml) are downloaded seperately. Since we are using Yarn in the new zimui, we can install these dependencies through Yarn and remove the openzim.toml file.

Dependencies installed using  `openzim.toml`:
- video-js
- chosen (Not required for new UI)
- jquery (Not required for new UI)
- ogvjs
- videojs-ogvjs
- webp-hero + polyfills

### LINKED ISSUE #230: Setup ogv.js as a fallback in video.js
The html5 player in `video.js` is currently used for video playback. However, to support older browsers like Safari for playing webm videos, we need to set up `Ogv.js` as a fallback.

However, there is an issue with integrating Ogv.js into the Vue.js app. In the [ogvs](https://github.com/openzim/youtube/tree/ogvjs) branch, I attempted to configure the [videojs-ogvjs](https://github.com/hartman/videojs-ogvjs) plugin, but encountered an error in the console:

```
TypeError: Cannot read properties of undefined (reading 'getComponent')
```

It seems that the videojs-ogvjs plugin is attempting to access video.js in a global scope but is unable to locate it.

dan-niles:
In this PR, I added ogv.js as a fallback to video-js for `webm` playback on unsupported browers. The changes are as follows:
- Setup videojs-ogvjs tech plugin in video-js (Custom script taken from [here](https://github.com/HuongNV13/videojs-ogvjs)) 
- Remove `openzim.toml` since all the dependencies for the UI are now managed using Yarn.
- Setup `vite-plugin-static-copy` to copy `ogvjs` files from `node_modules` into the final build/dist folder. (This is needed because the videojs-ogvjs tech plugin needs access to the ogvjs files at runtime)
- Add `type: "module"` back to package.json and rename `cypress.config.ts` to `cypress.config.js`.

(Tested `webm` playback on ogvjs by removing `html5` from the `TechOrder` option in video options.)

Close #230
Close #218


zimui/src/plugins/videojs-ogvjs.js
```diff
@@ -0,0 +1,759 @@
+/* eslint-disable no-prototype-builtins */
+/**
+ * videojs-ogvjs
+ * @version 1.0.0
+ * @copyright 2023 Huong Nguyen <huongnv13@gmail.com>
+ * @license MIT
+ */
+import videojs from 'video.js'
```
benoit74, line 8: Please add a comment explaining:
- where this file has been retrieved from 
- since if I'm not mistaken this is a Github repo, also specify commit ID + date when this was retrieved

Goal is that it is clear for maintainers how to update this file, and when this is needed

zimui/src/views/VideoPlayerView.vue
```diff
@@ -75,6 +75,10 @@ const videoURL = computed<string>(() => {
   return video.value?.videoPath || ''
 })
 
+const videoFormat = computed<string>(() => {
```
benoit74, line 78: Please add comment explaining this is a kind of trick to retrieve video format and add one or two sample values (I pretty sure it will always work, but should it become a problem it needs to be clear what this is doing)

benoit74 (changes requested):
Unfortunately I have a bug with kiwix-serve reader on Firefox 127.0.2 on Mac M1 with Sonoma 14.5.

image

But it works well on same machine with Chrome 125.0.6422.142 and with Safari 17.5 (19618.2.12.11.6).

Could you have a look if you achieve to reproduce the issue on your machine and/or if you find something which could explain how to solve this? This looks a bit like an edge-case where it tries to decode audio as mpeg instead of vorbis (not even sure what is the appropriate media type for vorbis), and for some reasons Firefox is less permissive than Chrome and Safari. For convenience the ZIM will soon be available at https://dev.library.kiwix.org/viewer#tests_eng_yt-zim_2024-07 (in more or less 10 to 20 minutes)

benoit74:
Other than this small glitch, very good work!

benoit74:
Nota: Firefox issue is in fact also visible with the https://pwa.kiwix.org and with the browser extension ; pwa works well on Chrome and Safari as well

zimui/src/plugins/videojs-ogvjs.js
dan-niles: Added in [832ac24](https://github.com/openzim/youtube/pull/259/commits/832ac248db071e4a381764721beb9b8049c0786c)

zimui/src/views/VideoPlayerView.vue
dan-niles: Added in [832ac24](https://github.com/openzim/youtube/pull/259/commits/832ac248db071e4a381764721beb9b8049c0786c)

dan-niles:
> Unfortunately I have a bug with kiwix-serve reader on Firefox 127.0.2 on Mac M1 with Sonoma 14.5.

@benoit74 I couldn't reproduce this issue. I tested on Firefox 127.0.2 on Mac M1 with Sonoma 14.5. It is working on kiwix-serve, https://pwa.kiwix.org/ and the Firefox extension for me and I didn't get the above mentioned console warnings.

I did however install a fresh copy of Firefox to do this testing. Can you check if any other extensions are causing conflicts?

https://github.com/openzim/youtube/assets/56271899/962c76d8-1984-4a5b-bf30-3bb749506426

benoit74:
I confirm that I cannot reproduce the issue on a Browserstack machine. Disabling all extensions does not help. But anyway, for now it really looks to me like a very local problem on my machine. Let's wait for the problem to appear on another machine (hopefully not) to confirm there is a real problem ; I suspect there is no problem. And anyway switching to another browser is also not a very big deal.

benoit74 (approved):
LGTM

PR #262, opened 2024-07-08
Add items to the ZIM "on the fly"

### LINKED ISSUE #204: Fail early if ZIM cannot be created at given location / name
Currently, if given output location or ZIM name are incorrect, we fail very late, after all stuff has been scraped.

We should fail way earlier.

We should use https://github.com/openzim/python-scraperlib/issues/163 (once implemented) for these checks.

Playlist mode where ZIM name is dynamically computed is probably going to be a bit more complex / not possible, but we will live without that (especially since anyway we do not use playlist mode in Zimfarm anymore)

### LINKED ISSUE #209: Use scraperlib & pylibzim properly - add items to the ZIM "on the fly"
We should move 
- from "zimwriterfs" mode where everything is stored on the disk and then ZIM is created from this on-disk structure 
- to adding items to the ZIM "on-the-fly" as their content is received.

dan-niles:
Close #204
Close #209

Changes in this PR:
- Remove `--no-zim` CLI argument.
- Add items "on the fly" to the ZIM:
   - Videos and video thumbnails are added to ZIM immediately after they are downloaded.
   - Channel branding of main channel (profile.jpg) is added after creating the favicon using it.
   - Subtitles are added to ZIM immediately after download. Information about language codes are cached so that it can be used later on to generate JSON files for individual videos.
 - Use `validate_zimfile_creatable` from `zimscraperlib` to validate if ZIM cannot be created at given location / name.
 - Remove `add_files_to_zim` method from Youtube2Zim class, since we no longer have to copy the whole build directory into the ZIM.

### DIFF
--- a/scraper/src/youtube2zim/scraper.py
+++ b/scraper/src/youtube2zim/scraper.py
@@ -10,7 +10,6 @@
 import concurrent.futures
 import datetime
 import functools
-import os
 import re
 import shutil
 import subprocess
@@ -277,9 +276,6 @@ class Youtube2Zim:
         # check that we can create a ZIM file in the output directory
         validate_zimfile_creatable(self.output_dir, self.fname)
 
-        # create output directory if it does not exist
-        os.makedirs(self.output_dir, exist_ok=True)
-
         # check that build_dir is correct
         if not self.build_dir.exists() or not self.build_dir.is_dir():
             raise OSError(f"Incorrect build_dir: {self.build_dir}")
@@ -363,6 +359,14 @@ class Youtube2Zim:
         self.zim_file.start()
 
         try:
+            logger.debug(f"Preparing zimfile at {self.zim_file.filename}")
+
+            logger.info("add main channel branding to ZIM")
+            self.add_main_channel_branding_to_zim()
+
+            logger.debug(f"add zimui files from {self.zimui_dist}")
+            self.add_zimui()
+
             # download videos (and recompress)
             logger.info(
                 "downloading all videos, subtitles and thumbnails "
@@ -390,13 +394,6 @@ class Youtube2Zim:
 
             logger.info("download all author's profile pictures")
             self.download_authors_branding()
-
-            logger.info("add main channel branding to ZIM")
-            self.add_main_channel_branding_to_zim()
-
-            logger.debug(f"Preparing zimfile at {self.zim_file.filename}")
-            logger.debug(f"Recursively adding files from {self.build_dir}")
-            self.add_zimui()
 
             logger.info("creating JSON files")
             self.make_json_files(succeeded)
--- a/scraper/src/youtube2zim/scraper.py
+++ b/scraper/src/youtube2zim/scraper.py
@@ -30,7 +30,7 @@ from zimscraperlib.image.transformation import resize_image
 from zimscraperlib.inputs import compute_descriptions
 from zimscraperlib.video.presets import VideoMp4Low, VideoWebmLow
 from zimscraperlib.zim import Creator
-from zimscraperlib.zim.filesystem import FileItem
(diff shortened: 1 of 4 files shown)


benoit74 (changes requested):
Good work! Caching subtitles to disk is indeed a promising compromise.

Aside discussions, I recommend to move as much as possible every steps before the long `download_video_files` step. I think that `download_authors_branding`, `add_main_channel_branding_to_zim` and `add_zimui` are good candidates. The idea is that since `download_video_files` takes lots of time, whatever can be run before that is good to run before, because should it fail, we prefer that it fail as soon as possible (and hence before the `download_video_files` step). Would it make sense?

benoit74:
Nota: just so that it is clear for our future-self, we removed the `--no-zim` argument since we are now adding files on-the-fly to the ZIM, so there is no added-value / sense to ask to not create the ZIM.

benoit74:
Oh and btw, we should remove as well the `keep_build_dir` / `--keep` as well, build_dir content is not interesting anymore to be kept.

And the documentation about `--tmp-dir` argument should be revisited, it is not really accurate anymore.

dan-niles:
> I recommend to move as much as possible every steps before the long `download_video_files` step. I think that `download_authors_branding`, `add_main_channel_branding_to_zim` and `add_zimui` are good candidates.

I have moved `add_main_channel_branding_to_zim` and `add_zimui` to run before the download step in [869b499](https://github.com/openzim/youtube/pull/262/commits/869b499446bc736cd6b0bceb979b4e334279bc9c). However `download_authors_branding` cannot be moved since it depends on the `get_videos_authors_info` method running before it. 

`get_videos_authors_info` fetches all the information about the authors of videos and then `download_authors_branding` downloads the branding for the fetched authors.

dan-niles:
`--keep` CLI argument removed in [57f5a5d](https://github.com/openzim/youtube/pull/262/commits/57f5a5d1fc1726e09db29d48b9ad8077d491fa29). I also updated the documentation for `--tmp-dir` in the same commit.

scraper/src/youtube2zim/scraper.py
```diff
@@ -259,6 +265,25 @@ def run(self):
...
+            self.fname.format(period=period)
+            if self.fname
+            else f"{self.name}_{period}.zim"
+        )
+
+        # check that we can create a ZIM file in the output directory
+        validate_zimfile_creatable(self.output_dir, self.fname)
+
+        # create output directory if it does not exist
+        os.makedirs(self.output_dir, exist_ok=True)
```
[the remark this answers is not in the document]
dan-niles: Removed in [869b499](https://github.com/openzim/youtube/pull/262/commits/869b499446bc736cd6b0bceb979b4e334279bc9c)

scraper/src/youtube2zim/scraper.py
```diff
@@ -412,10 +407,6 @@ def run(self):
             logger.info("Finishing ZIM file…")
             self.zim_file.finish()
 
-        if not self.keep_build_dir:
```
benoit74: By default, `keep_build_dir` was False, so we want to keep the two lines below.

benoit74 (changes requested):
One more thing

scraper/src/youtube2zim/scraper.py
benoit74: You can amend commit 57f5a5d1fc1726e09db29d48b9ad8077d491fa29 directly

dan-niles: Oops 😅 Fixed it in [1815f28](https://github.com/openzim/youtube/pull/262/commits/1815f280de9a13466b204f4d3015850a866747bf)

benoit74 (approved):
LGTM

PR #263, opened 2024-07-10
Add support for color CLI arguments in new UI

### LINKED ISSUE #217: Add support for color CLI arguments to work with the new UI
With the introduction of the new UI, we need to support the arguments in the [entrypoint.py](https://github.com/openzim/youtube/blob/main/scraper/src/youtube2zim/entrypoint.py) file that alter the look of a channel.
- [ ] --main-color (Sets the main color for the channel)
- [ ] --secondary-color (Sets the secondary color for the channel)

dan-niles:
Close #217

Changes:
- Add `Config` schema and save `main-color` and `secondary-color` in config.json
-  Update Vuetify initialization to load main and secondary colors from config.json
- Update UI to support different background colors from config.json
  - `secondary-color` is used as the background for the ZIM UI.
  - `main-color` is used for the banner if a banner image is not available.

Screenshots:
image
image


scraper/src/youtube2zim/schemas.py
```diff
@@ -100,3 +100,10 @@ class Channel(CamelModel):
     joined_date: str
     collection_type: str
     main_playlist: str | None = None
+
+
+class Config(CamelModel):
+    """Class to serialize configuration data for the ZIM."""
```
benoit74, line 106: "for the ZIM." => "for the ZIM UI." ?

scraper/src/youtube2zim/schemas.py
```diff
@@ -100,3 +100,10 @@ class Channel(CamelModel):
...
+
+    main_color: str | None = "#000000"
+    secondary_color: str | None = "#FFFFFF"
```
benoit74, line 109: I think we should remove these two default values, they are never used in current codebase and could mislead a maintainer in believing that they are the default when no CLI argument is passed, while the default is in JS code.

benoit74 (changes requested):
Good work !

I like the fact that this PR introduces many useful things in terms of architecture:
- a config file to store whatever configuration values is needed to customize the UI for some ZIM
- the `loadVuetify` method and the Vuetify theme, to load the config early and be able to customize the theme early, so that we do not have strange side-effects like flickering

Only two very small comments.

scraper/src/youtube2zim/schemas.py
dan-niles: Changed in [35c4e76](https://github.com/openzim/youtube/pull/263/commits/35c4e767b329aacee1f51d67e8f3b15196163f97)

scraper/src/youtube2zim/schemas.py
dan-niles: Removed in [35c4e76](https://github.com/openzim/youtube/pull/263/commits/35c4e767b329aacee1f51d67e8f3b15196163f97)

benoit74 (approved):
LGTM, thank you

PR #265, opened 2024-07-11
Restore search functionality by indexing content in JSON files

### LINKED ISSUE #224: UI Revamp: restore search functionality
In every ZIM, we need to have proper information in the two search indexes which are inside the ZIM so that suggestion search and full-text search are working properly in readers supporting these indexes.

Due to the move to Vue.JS in the UI Revamp, we do not have anymore one HTML file per video ; search indexes are hence not updated anymore, and search is just not working anymore.

We need to restore the search functionality by adding proper information in the two search indexes. 

Nota: in former UI, suggestions and full-text search were only returning videos, not playlists, let's start with this for now.

Adding this search functionality in a Vue.JS-based scraper has been already done in freecodecamp: https://github.com/openzim/freecodecamp/blob/a1221073049895609d7712a8834474311535f0c1/scraper/src/fcc2zim/build.py#L68-L88

The idea is to create "fake" HTML pages which use the `<meta http-equiv="refresh" content="0;URL='xxxx'" />` functionality to immediately redirect the user to the proper page. This allows to set proper title in the search index for suggestions. And this could be used to trick the search index by storing whatever we want in full-text search in the HTML body.

This is however clearly a bit tricky (create a fake HTML entry where we do not want to generate HTML anymore ... well done ...) + a waste of space (we store for instance the video description twice, once in the fake HTML and once in the JSON used by Vue.JS UI)

@rgaudin as far as I remember, we spoke about this same need for Kolibri, and as far as I remembered you mentioned a solution to have this full-text search working as well without having to store the content in a real HTML file but just by updating the search indexes appropriately; of course I do not find the notes I've made at that time, probably because I forgot to write them down in a Github issue... sorry about that ; would you mind to point us in the right direction again. I remember about using a ZIM redirect entry (with `add_redirect`), but I don't see now how it could work so I might be totally mistaken ^^ Or maybe these were more ideas for an issue to open at libzim level so that we can specify title and content, and redirect to a path including a fragment?

dan-niles:
Changes:
- Add support to index content from `zimui` JSON files in the ZIM using custom `IndexData`
- Index title and description of all videos and playlists

Close #224

### DIFF
--- a/scraper/pyproject.toml
+++ b/scraper/pyproject.toml
@@ -31,7 +31,7 @@ additional-keywords = ["youtube"]
 [project.optional-dependencies]
 scripts = ["invoke==2.2.0"]
 lint = ["black==24.4.2", "ruff==0.4.10"]
-check = ["pyright==1.1.368"]
+check = ["pyright==1.1.371"]
 test = ["pytest==8.2.2", "coverage==7.5.3"]
 dev = [
   "pre-commit==3.7.1",
--- a/scraper/src/youtube2zim/scraper.py
+++ b/scraper/src/youtube2zim/scraper.py
@@ -21,6 +21,7 @@ from typing import Any
 
 import yt_dlp
 from kiwixstorage import KiwixStorage
+from libzim.writer import IndexData  # type: ignore
 from pif import get_public_ip
 from zimscraperlib.download import stream_file
 from zimscraperlib.filesystem import delete_callback
@@ -31,7 +32,7 @@ from zimscraperlib.image.probing import get_colors, is_hex_color
 from zimscraperlib.image.transformation import resize_image
 from zimscraperlib.inputs import compute_descriptions
 from zimscraperlib.video.presets import VideoMp4Low, VideoWebmLow
-from zimscraperlib.zim import Creator
+from zimscraperlib.zim import Creator, StaticItem
 from zimscraperlib.zim.filesystem import validate_zimfile_creatable
 from zimscraperlib.zim.metadata import (
     validate_description,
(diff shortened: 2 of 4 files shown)


scraper/src/youtube2zim/scraper.py
```diff
@@ -84,6 +85,29 @@
...
+        return self.title
+
+    def get_content(self):
+        return self.content
+
+    def get_keywords(self):
+        return ""
+
+    def get_wordcount(self):
+        return 1
```
benoit74, line 108: You should compute the number of words, probably `return len(self.content.split()) if self.content else 0` would do the job

scraper/pyproject.toml
```diff
@@ -31,7 +31,7 @@ additional-keywords = ["youtube"]
 [project.optional-dependencies]
 scripts = ["invoke==2.2.0"]
 lint = ["black==24.4.2", "ruff==0.4.10"]
-check = ["pyright==1.1.368"]
+check = ["pyright==1.1.371"]
```
benoit74, line 34: If you update this dependency, you should align the version in `.pre-commit-config.yaml`

benoit74 (changes requested):
Well done ! FYI, I'm currently embedding the logic around the custom index data in a "simpler" API in python-scraperlib (https://github.com/openzim/python-scraperlib/pull/182, not yet ready), at some point when this will be released we will use this new scraperlib functionality to simply youtube scraper maintenance. Until then, your code is OK.

Two small things

scraper/pyproject.toml
dan-niles: Fixed it in https://github.com/openzim/youtube/pull/265/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R34

scraper/src/youtube2zim/scraper.py
dan-niles: Fixed in https://github.com/openzim/youtube/pull/265/files#diff-ba8e4270b357604b029038e7742f741b431b493a07d7482ea99e22b7002388cdR108

benoit74 (approved):
All good, thank you!

Issue #261, opened 2024-07-08
Test the new scraper on existing Zimfarm recipes


dan-niles:
Up to now, we've been developing the scraper and testing it with the [openZIM_testing](https://www.youtube.com/channel/UC8elThf5TGMpQfQc_VE917Q) YouTube channel. It would be beneficial if we could test it on actual Zimfarm recipes to ensure the new scraper meets all the requirements.

@Popolechien Could you share a few (2-5) Zimfarm recipes to test out the new YouTube scraper? I believe it would be helpful to test and ensure everything is working correctly with the new scraper.

benoit74:
I suggest to start with [deus-ex-silicium_fr_all](https://farm.openzim.org/recipes/deus-ex-silicium_fr_all) which has just been rebuilt. This is a significantly big channel, with multiple playlists and where current UI is clearly clumsy.

We should probably simply reconfigure the recipe to use `dev` Docker image and push to dev library.

Popolechien:
The Canadian prepper playlists are good ones to try our hand at: 
https://farm.openzim.org/recipes?tag=youtube&name=canadian_prepper

benoit74:
@RavanJAltaie @Popolechien 
I'll transition https://farm.openzim.org/recipes/deus-ex-silicium_fr_all and https://farm.openzim.org/recipes/canadian_prepper_bugoutroll_en to the `dev` image and pushing to the `dev` library to test the new ZIM UI.
I will do the same with other canadian prepper once titles have been fixed and a new ZIM has been produced with the old UI (see my message on Slack).
Please do not touch anymore these recipes until further notice (and do not use these recipes as clone source when cloning a recipe to create a new Youtube recipe, since the new Youtube UI is not yet ready for production, so you shouldn't publish ZIM from `dev` image to production).

benoit74:
Two first ZIMs have been produced today:
- https://dev.library.kiwix.org/viewer#canadian_prepper_bugoutroll_en_2024-07
- https://dev.library.kiwix.org/viewer#deus-ex-silicium_fr_all_2024-07

Popolechien:
This looks very, very good.

Question: the Canadian Prepper zim is that of a single playlist. Shouldn't the playlist tab be disabled in that case?
Also, I see the banner is missing in CP, and replaced by a grey-black one for Deus ex. Bug or feature?

Also: `About channel` -> `About`or `About this channel`

dan-niles:
@Popolechien 
[...]

PR #266, opened 2024-07-12
Hide playlists tab in `zimui` if there is only 1 playlist

dan-niles:
As per the suggestion in https://github.com/openzim/youtube/issues/261#issuecomment-2223121230, the `Playlist` tab in the Vue.js UI can be hidden if there is only 1 playlist in the ZIM. Changes in this PR:
- Add `playlist_count` to channel.json
- Modify `ChannerHeader.vue` to hide the tabs based on the `playlist_count` value.

Screenshot (ZIM with only 1 playlist):
image


benoit74 (approved):
LGTM

PR #267, opened 2024-07-15
Fix channel title truncation for smaller screens in zimui

### LINKED ISSUE #264: Channel title in video player page breaks into multiple lines when too long
The channel title in video player page breaks into multiple lines when it is too long. The title should be truncated on smaller screens.

image

dan-niles:
Fix #264

Fix the channel title truncation on smaller screen devices, so that the channel name doesn't break into multiple lines.

Before:
image

After:
image


benoit74 (approved):
I like the "small" about button as well, good point, it was a bit "clunky" as well.

PR #272, opened 2024-07-17
Update Vite configuration in zimui for native ES6 support

### LINKED ISSUE #243: New UI: add support for old browsers
We need to consider supporting more browsers than are supported by default in Vite.

According to https://github.com/vitejs/vite/tree/main/packages/plugin-legacy, Vite's default browser support baseline is [Native ESM](https://caniuse.com/es6-module), [native ESM dynamic import](https://caniuse.com/es6-module-dynamic-import), and [import.meta](https://caniuse.com/mdn-javascript_operators_import_meta).

Vue.JS 3 [baseline](https://vuejs.org/about/faq#what-browsers-does-vue-support) is [browsers with native ES2015 support](https://caniuse.com/es6).

I suggest that we consider covering all browsers supported by Vue 3, by adapting Vite configuration. Proper list seems to be `fully supports es6`: https://browsersl.ist/#q=fully+supports+es6, meaning:
- Chrome 51 released in 2016
- Edge 15 released in 2017
- Firefox 54 released in 2017
- Opera 38 released in 2016
- Safari 10 on macOS/iOS released in 2016
- Not sure about browser for phones, data on caniuse.com seems erroneous/incomplete

Sample configs for vite-plugin-legacy are available at https://github.com/vitejs/vite/discussions/10519

dan-niles:
Close #243

Added support for [browsers with native ES6 support](https://caniuse.com/es6) by implementing `@vitejs/plugin-legacy` with `fully supports es6` as the target setting.

I was unable to test this on older browsers. This should be tested on the lowest browser versions available on https://browsersl.ist/#q=fully+supports+es6


benoit74 (changes requested):
This is unfortunately not working on old browsers.

I tried to experiment a bit by setting `renderLegacyChuncks: true` and `modernPolyfills: false` and removing `build - target`. It seems to progress a little bit more, but still doesn't achieve to work. It stops on a ResizeObserver error. 

It looks like configuration which is done will polyfill vite functionalities, but not other functionalities that we've used and needs to be polyfilled as well.

It made me realize that Vuetify 3 is not supporting such old browser and explicitely recommend to add polyfills: https://vuetifyjs.com/en/getting-started/browser-support/#browsers

Let's sync via chat on how to move this forward.

dan-niles:
I was able to get the UI working on Chrome 60 and above by removing the `esbuild` and `build` properties from the Vite config and using `resize-observer-polyfill`. The `esbuild` property was causing a conflict with `vite-plugin-legacy` which automatically detects and applies the necessary polyfills. Once esbuild was removed, the proper polyfills were included in the final build.

I also made some slight tweaks to the UI in [37e6c93](https://github.com/openzim/youtube/pull/272/commits/37e6c93d4780f840f40301a55bf926d1a4950d69), ensuring the UI renders properly on older browsers.

Some issues that are present still are:
- The polyfill for `webp` images applied using `webp-hero` is not working (tested on Firefox 54).
- For Chrome versions 51-59, there seems to be an issue with `video-js`. I tried removing all video-js imports, and the UI works fine on these browsers. I suspect there might be something wrong since we are using the latest version of `video-js`.

benoit74:
This is already a significant step forward ! Keep going ^^

dan-niles:
Fixed the polyfill issue for `webp` images in [04927e0](https://github.com/openzim/youtube/pull/272/commits/04927e09413e0dab45d4aa0c2b76e60b74bc0080). I tested on Firefox 64 and everything is working fine. 

However, there is a bug in Firefox 54 where the thumbnail height is not set properly (See screenshot below). The polyfilled image is rendered in the DOM but there is an issue with the height. This is probably a CSS bug, I'll check this and update this PR.

image

dan-niles:
The fix for the height issue was pretty simple ([a963b84](https://github.com/openzim/youtube/pull/272/commits/a963b84d76ea25853f39be2cf91a8ae300ee7787)). Just had to change the display type of the thumbnails from `flex` to `block`.

dan-niles:
With [d9007b3](https://github.com/openzim/youtube/pull/272/commits/d9007b32400588a04d9b71f5f0e3880f58530cff), the UI and video playback now works properly on `Firefox 54 and above`. In this update, I added the source and subtitle tracks within the `<video>` tag in the `VideoPlayer.vue` component. If `video.js` fails to work on older browsers, it will fallback to the native browser video playback.

However, there is still an issue with video playback in `Chrome 51-59`, which seems to be related to `video.js `v8` (latest), specifically the minified bundle used in the final build on these Chrome versions (see [video.js issue #8783](https://github.com/videojs/video.js/issues/8783)).

I am currently checking the possibility of using the un-minified version of video.js in our final build. If this is not possible, we might need to consider downgrading to video.js `v7`. Given that video.js `v7` was released nearly two years ago, I am not sure if downgrading is a worthwhile tradeoff just to support these older browsers.

dan-niles:
I had to downgrade `ogv.js` from `v1.9.0` to `v1.8.9` (Same version used in the [V2 scraper](https://github.com/openzim/youtube/blob/63dc65e348d1429d8424398a3236e3e2987ffd98/openzim.toml#L26)) in [64735f5](https://github.com/openzim/youtube/pull/272/commits/64735f505de4c2f7904995aa6327d7e5de5fbd8f), to get the video playback working on older Safari browsers. 

There was a small issue in Safari 11.1 where the tab switcher in the channel header wouldn't work. I wrapped the `tab` component with a `router-link` in [6b70dba](https://github.com/openzim/youtube/pull/272/commits/6b70dba722fae49a2a1bfc012c1a84d0f4bbccd4) and fixed it.

Everything works fine on `Safari 11.1 and above`. 

However, Safari 10.1 throws the following error:
image
This needs to be checked.

dan-niles:
> However, there is still an issue with video playback in `Chrome 51-59`, which seems to be related to video.js `v8` (latest), specifically the minified bundle used in the final build on these Chrome versions.

A PR (https://github.com/videojs/video.js/pull/8810) has been opened and merged in video.js to patch this issue. Once they make a release with this patch, we can test it out and see if it resolves our issue.

zimui/package.json
```diff
@@ -42,8 +44,9 @@
     "eslint-plugin-vue": "^9.23.0",
     "jsdom": "^24.0.0",
     "npm-run-all2": "^6.1.2",
-    "ogv": "^1.9.0",
+    "ogv": "1.8.9",
```
benoit74, line 47: Please add a comment with a link to PR comment explaining why we pinned this version.

zimui/src/plugins/vuetify.ts
```diff
@@ -9,9 +10,9 @@ async function loadVuetify() {
 
   // Load primary and secondary colors from config.json
   try {
-    const response = await fetch('./config.json')
-    if (response.ok) {
-      const config: Config = await response.json()
+    const response = await axios.get('./config.json')
+    if (response.status === 200) {
```
benoit74, line 14: Replace `200` by `HttpStatusCode.Ok`, with `import { HttpStatusCode } from axios`

zimui/src/plugins/webp-hero.ts
```diff
@@ -1,11 +1,26 @@
...
-    if (!support_webp) {
-      const webpMachine = new WebpMachine()
-      webpMachine.polyfillDocument()
+const webp = new Webp()
+
+export async function polyfillThumbnail(src: string): Promise<string> {
+  const isWebpSupport = await detectWebpSupport()
+  try {
+    if (!isWebpSupport) {
+      const wM = new WebpMachine({ webp })
```
benoit74, line 17: `webpMachine` was way clearer than `wM` from my perspective. Any reason to prefer `wM`?

benoit74 (changes requested):
Few small changes, thank you.

zimui/package.json
dan-niles: Comments are not allowed in JSON files. Do you have any suggestions where else I can mention this?

benoit74: Damn it ; let's live without it then

zimui/src/plugins/vuetify.ts
dan-niles: Changed in [664875a](https://github.com/openzim/youtube/pull/272/commits/664875ab24710957d5240d3a12f786544b72ad70)

zimui/src/plugins/webp-hero.ts
dan-niles: Changed in [82feca5](https://github.com/openzim/youtube/pull/272/commits/82feca5fdf816dfdf510da9b3a08eae589d2409a)

benoit74 (commented):
LGTM, please rewrite commit history (at least two last commits should not be part of the history now that review is completed) and I will merge once done.

dan-niles:
Sqashed and rewrote commit history

benoit74 (approved):
LGTM, thank you

PR #280, opened 2024-07-25
Add an overlay image for the channel banner

### LINKED ISSUE #279: Add nice overlay on banner
When no banner image is used, only a single color is used. When color is white or close to white, it makes a weird gradient. 

image

In some old browsers, the gradient does not even load.

It would be nice to have some kind of overlay pattern to make the banner visible in all situations and limit the "weirdness" of this part of the UI under all circumstances.

dan-niles:
Add an overlay image for the channel banner. As discussed in #279 I added the "ZigZag" pattern overlay.

image

Tested on Chrome 60, Firefox 54 and Safari 10.1
Close #279


benoit74 (approved):
LGTM, I just rewrote the commit to add a CHANGELOG entry so that I can merge (and most probably release) right away

PR #283, opened 2024-08-02
Disable preloading of subtitles in video.js

### LINKED ISSUE #38: Do not preload subtitles in video.JS
When building ZIM with `--all-subtitles`, all videos gets more than a hundred subtitles. Subtitles are loaded asynchronously on page load but the video and controls only starts/become active when all of them have been called which can takes seconds.

dan-niles:
Close #38

This PR disables the preloading of subtitles in video.js for `zimui`. Changes include:
- Set `preloadTextTracks` as false for both `html5` and `ogvjs` tech plugins.

Tested on `Chrome v127` and `Safari 11.1`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Disable preloading of subtitles in video.js in `zimui` (#38)
 
 ## [3.0.0] - 2024-07-29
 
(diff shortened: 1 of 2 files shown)


benoit74 (changes requested):
Please add a mention that we patched this file in the comment block where we referenced the source and commit-id, with a link to this specific commit, otherwise it will get lost next time we update / it will be hard to remember what needs to be patched.

Please also open an issue + a PR at https://github.com/HuongNV13/videojs-ogvjs so that we have a chance this is merged upstream + we share this with "the world".

dan-niles:
I made a mistake here. It looks like we don't need to edit the `videojs-ogvjs` plugin. The plugin already extends the [Tech](https://github.com/openzim/youtube/blob/402bf3722e353aced3c2412d790d72a80c110d61/zimui/src/plugins/videojs-ogvjs.js#L79) class of video.js which means we can directly use the `preloadTextTracks` option for `ogvjs` without any modifications to the plugin.

benoit74 (approved):
LGTM, thank you!

PR #287, opened 2024-08-04
Cache subtitles in S3 storage

### LINKED ISSUE #277: Cache subtitles on S3 as well
Currently only video thumbnails and video themselves are cached on S3.

This has the drawback that when an IP has been blacklisted from yt-dlp usage, the recipe fails to produce the ZIM even if all API calls have succeeded, because we use yt-dlp to download the subtitles.

Caching the subtitles on S3 would allow to create the ZIM.

dan-niles:
This PR modifies the `download_subtitles` method to cache subtitles in S3. The modified method now works as follows:

- Fetch requested subtitle keys using `yt-dlp` (e.g., `en`, `fr`, `de`) and store them in `requested_subtitle_keys`.
- Iterate through the `requested_subtitle_keys` and attempt to download each subtitle file from the S3 cache. If a file is - successfully downloaded from the S3 cache, remove the corresponding key from `requested_subtitle_keys`.
- For the remaining keys in `requested_subtitle_keys`, download the subtitles using `yt-dlp`.
- Upload the newly downloaded subtitles to the S3 cache.
- Save the information about the fetched subtitles in a local cache as a JSON file for future use.
- Add the downloaded subtitles to the ZIM file.

Close #277

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Disable preloading of subtitles in video.js in `zimui` (#38)
+- Update `download_subtitles` method to cache subtitles in S3 storage (#277)
 
 ## [3.0.0] - 2024-07-29
 
--- a/scraper/src/youtube2zim/scraper.py
+++ b/scraper/src/youtube2zim/scraper.py
@@ -882,21 +882,70 @@ class Youtube2Zim:
     def download_subtitles(self, video_id, options):
         """download subtitles for a video"""
 
+        def get_subtitle_s3_key(code: str) -> str:
+            return f"subtitles/{video_id}/subtitle.{code}.vtt"
+
+        def get_subtitle_path(code: str) -> str:
+            return options["y2z_videos_dir"].joinpath(f"{video_id}/video.{code}.vtt")
+
         options_copy = options.copy()
-        options_copy.update({"skip_download": True, "writethumbnail": False})
+        options_copy.update(
+            {"skip_download": True, "writethumbnail": False, "listsubs": True}
+        )
+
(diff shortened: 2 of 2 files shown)


benoit74 (commented):
I'm a bit puzzled about this issue/PR now.

Original goal I expressed in the issue was to avoid being blocked by yt-dlp ban when subtitles did not changed. Goal is not reached in this PR since we still need to get the list of subtitles with yt-dlp.

At the same time, caching as well the list of subtitles is probably not something wishable, we usually do not cache in S3 the responses to API calls, but rather resources which take time/resources to recompute (reencoded videos / images). The idea of caching subtitles in S3 was probably already a deviation from this usual behavior.

To help sort this out, can you please share some data about how this change makes the scraper run faster or not?

dan-niles:
> Original goal I expressed in the issue was to avoid being blocked by yt-dlp ban when subtitles did not changed. 

The reason I used yt-dlp to get the list of subtitles is that the scraper doesn't know which language subtitles need to be downloaded. Without the `--all-subtitles` flag, only the default available subtitles for the video are downloaded. With the `--all-subtitles` flag, auto-generated subtitles are included as well.

To avoid calling yt-dlp entirely in this scenario, we could save two zipped files in the S3 cache:
1. `subtitles/{video_id}/default.zip` - containing only the default subtitles.
2. `subtitles/{video_id}/all.zip` - containing all subtitles when `--all-subtitles` is passed.

WDYT?

benoit74:
This seems a potential improvement, but does it really help the scraper run faster? Because it has the drawback that we do not know when we should invalidate these to update them.

benoit74:
Let's pause this issue/PR to let me reflect a bit on this

benoit74:
Superseeded by https://github.com/openzim/youtube/pull/460, thank you @dan-niles anyway

PR #301, opened 2024-08-30
Use multithreaded YouTube downloader from zimscraperlib

### LINKED ISSUE #118: Use newer multithreaded youtube downloader from scraperlib
We now have support for downloadeing youtube videos on a fixed number of threads in scraperlib. Hence, we shall use that instead of the current approach of downloading in batches

dan-niles:
Close #118

Use newer [multithreaded YouTube downloader](https://github.com/openzim/python-scraperlib/blob/335d5271e106b374f1aca871d19557ff2c81582d/src/zimscraperlib/download.py) from `zimscraperlib` to download videos, thumbnails and subtitles. 

Changes:
- Remove "batch" downloading of videos and `download_video_files_batch` method.
- Update `download_video`, `download_thumbnail` and `download_subtitle` methods to use new multithreaded downloader from scraperlib.
- Update the above methods to download the files asynchronously.


rgaudin:
Changed reviewer as @benoit74 is coming back

benoit74 (changes requested):
This is not working exactly as intended, when a video fails to download, it is not counted as failed and final ZIM is hence created with no video at all (in my test where I not-so-intentionally had an issue with youtube-dlp).

`download_video` and `download_thumbnail` are now returning a future, so one has to check its status after completion (not an easy feat tbh).

dan-niles:
@benoit74 I've updated the code to handle errors correctly. It now properly returns both the `succeeded` and `failed` video_ids. I've also mentioned the time taken to create a ZIM of the `openZIM_testing` channel with different values for `concurrency` (without using s3 cache) below.

| Concurrency | Time (s) |
| ----------- | -------- |
| 1           | 40.16    |
| 2           | 25.29    |
| 4           | 21.11    |

However, I realized that most ZIMs in zimfarm are downloaded using the S3 cache. With this approach, only the videos that are not in the S3 cache and are downloaded directly using `yt-dlp` will benefit from the parallelization. This is because I'm using the [ThreadPoolExecutor](https://github.com/openzim/python-scraperlib/blob/335d5271e106b374f1aca871d19557ff2c81582d/src/zimscraperlib/download.py#L27) in `YoutubeDownloader` from `python-scraperlib`, rather than downloading in batches as suggested in https://github.com/openzim/youtube/issues/118#issue-700516092. Therefore, this change will only provide a performance gain in cases where videos are missing from the S3 cache.

Given that most downloads rely on the S3 cache, would it be better to keep the current approach of downloading in batches and using the `YoutubeDownloader` from scraperlib without concurrent execution?

WDYT?

benoit74:
> Given that most downloads rely on the S3 cache, would it be better to keep the current approach of downloading in batches and using the YoutubeDownloader from scraperlib without concurrent execution?

This is indeed a very important point! Current implementation allow concurrent download of both S3 cached videos and Youtube videos, and also reencoding.

I'm sorry to not have noticed this before. I feel a bit ashamed to have wasted your time on this. Thank you for speaking up, it would have been a bad move indeed!

I propose to close the issue and the PR as "non-relevant", WDYT? Or at least we should enhance the scraperlib, but this is not a small change to implement.

PR #306, opened 2024-09-03
Upgrade dependencies, including zimscraperlib 4

benoit74:
Most commits have been suggested by dependabot, I just rebased them in on single branch and added needed adaptations in few additional commits.

Nota: eslint has been rolled-back to 8.57.0, because upgrade to 9 is huge and I failed to make it work properly (same problem encountered on zimit-frontend, looks like Vue.JS + typescript within Vue.JS files is not a common combination in eslint users ...)


benoit74:
Remark: current version of https://github.com/vuejs/create-vue is still using eslint 8.57.0, so it is probably a good indicator that the ecosystem is really not yet ready, it is not just me and my limited JS knowledge. See https://github.com/vuejs/create-vue/issues/488

rgaudin (approved):
LGTM

PR #339, opened 2024-09-20
Merge behaviors of user/channel types and add support for `forHandle`

### LINKED ISSUE #338: Impossible to scrape channel Quickguidesformedicine
It is impossible to scrape channel https://www.youtube.com/@Quickguidesformedicine

Using any combination of `Quickguidesformedicine` or `UC1gMw6SLKx3t7hSF8s20JsQ` with type `user` or `channel` is not working, channel is never found.

benoit74:
Fix #338 (and make life easier for everyone)

Rationale:
- it is cumbersome for the end-user to try both `channel` and `user` types, there is no way to know in advance which one is correct
- it is easy to programmatically test both parameters one after the other one, there is probably no conflict possible
- since early 2024, Youtube API supports a new `forHandle` parameter which is probably the one we always want to use (but we have recipes which might still be configured by technical ID, so let's keep previous criteria as a fallback)
- I do not want to wait for a major release to drop `--type user` so for now I left it as accepted but merge behavior, and a warning is displayed about the fact that this is a deprecated value


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Merge behaviors of user/channel types and add support for `forHandle` (#339, fix for #338)
+  - `--type user` is now **DEPRECATED**
```
rgaudin, line 12: Remember the empty line after the header.

I believe a `### Deprecated` section would be more visible and practical.

rgaudin (approved):
👍

PR #349, opened 2024-10-07
Update layout of `Videos` tab in `zimui` to display videos from all playlists in the ZIM

### LINKED ISSUE #337: UI is not always really relevant when many playlists are in the ZIM
We have two tabs in Youtube UI front page:
- Videos
- Playlists

By default, the videos tab is selected

The videos tab show videos from the default playlist (first one of the list, first one of the channel, ...)

This means that by default when many playlists are present, we see only a subset of all videos. This is not really relevant / confusing.

See e.g. https://library.kiwix.org/viewer#s2underground_en_all . In this ZIM, at first sight, one can believe there is only 35 videos in the ZIM, because by default the "Long-Form Content" playlist has been selected.

I see multiple solutions forward:
- (1) drop the distinction between videos and playlists ; when there is only one playlist or no playlist at all, display the videos inside the playlist ; otherwise, display the list of playlists (maybe with a design where we see thumbnails from first videos of every playlist, like below - taken from Youtube)

image

- (2) by default, when there is multiple playlists, remove the "Videos" tab from UI and only show the "Playlists" one.

While 2 is obviously way easier to implement, I believe that only something like 1 make real sense.

dan-niles:
Fix #337

Update layout of `Videos` tab in `zimui` to display videos from all playlists in the ZIM. The videos from each playlist are now listed in separate rows as carousels.

Screenshots:
image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Merge behaviors of user/channel types and add support for `forHandle` (#339, fix for #338)
+- Update layout of `Videos` tab in `zimui` to display videos from all playlists in the ZIM (#337)
 
 ## [3.1.0] - 2024-09-05
 
--- a/scraper/src/youtube2zim/schemas.py
+++ b/scraper/src/youtube2zim/schemas.py
@@ -61,6 +61,7 @@ class Playlist(CamelModel):
     """Class to serialize data about a YouTube playlist."""
 
     id: str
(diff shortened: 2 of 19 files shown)


benoit74 (approved):
Impressive work, thanks a lot!

I do not have much to say, it looks like you spend quite a lot of time testing this and it is working very well.

My only concern is about the two tabs we have on the main page. I'm not sure they are really useful anymore. But I will open a dedicated issue on that, since this needs to be answered by more people than the two of us, and having ZIMs to showcase the situation is probably going to help.

 I will rebase and merge right away.

PR #365, opened 2024-10-14
Raise exception if there are no videos in the playlists

### LINKED ISSUE #347: What happens when there is no video in channel / playlist?
From https://github.com/openzim/youtube/pull/341#pullrequestreview-2324996456

What happens when the channel has no video, or the playlist(s) have no video in the end? Do we create an almost empty ZIM?

We need to ensure that scraper fails in such a situation with a clear error message.

dan-niles:
Fix #347

Raise exception if there are no videos in the playlists instead of creating an empty ZIM file. 
Tested with this playlist: https://www.youtube.com/playlist?list=PL5glJe_RySVnGr4P1ZCW3Ou32pkz0z-yY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


benoit74 (approved):
LGTM

PR #370, opened 2024-10-21
Use proper ZIM metadata key for `Scraper` and `Tags`

### LINKED ISSUE #369: [REGRESSION] Tags and Scraper metadata is incorrectly populated
See https://github.com/kiwix/operations/issues/286

benoit74:
Fix #369 

Changes:
- I've also enhanced a bit the integration tests for better checks (even if we can see that we already had a test about `scraper` metadata instead of `Scraper`, so it will not solve the problem we had in this issue at all)


rgaudin (approved):
👍

PR #372, opened 2024-10-31
Simplify scraper by dropping `--type` CLI argument

### LINKED ISSUE #361: Drop `--type` setting
We recently merged `--type channel` and `--type user`. I think we should go even further and simply remove the `--type` setting.

It means we currently have either `--type channel` or `--type playlist`. This `--type` parameter is however mostly useless because the ID we are passing are unique, we cannot confuse a playlist with a channel or a user or a handle.

It looks like it is often complex for the user to properly set this setting.

I think we could hence simply drop this `--type` setting and try to guess which type of `--id` has been passed, we could always consider the ID passed with `--id` can be a channel, a user, a handle or a playlist, and test the 4 possibilities in a row.

This would help save lots of headaches to our users and  it takes very little time to test the 4 possibilities in a row.

For backward compatibility reasons, the `--type` parameter should be kept until next major, but should simply be ignored,

I didn't check how easy this is to implement (we have maybe special flow when `--type playlist` is used).

benoit74:
Fix #361

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -100,7 +100,7 @@ jobs:
         env:
           YOUTUBE_API_KEY: ${{ secrets.YOUTUBE_API_KEY }}
           OPTIMIZATION_CACHE_URL: ${{ secrets.OPTIMIZATION_CACHE_URL }}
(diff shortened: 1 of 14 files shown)


rgaudin (approved):
LGTM

PR #388, opened 2025-03-12
Feat: support video chapters

### LINKED ISSUE #170: Add support for chapters
Teachers & Learners need clickable Tables-of-Contents to navigate long videos, to greatly deepen learning around "2-hour movies" and similar long videos.

How should this thoughtful interactivity be implemented in a ZIM file, for long videos especially?

YouTube calls this [Video Chapters](https://support.google.com/youtube/answer/9884579).  Here's a great example beginning with an 11-hour YouTube video on Learning CSS:

image

Above screenshot is from the Description > "Show more" section of https://youtu.be/OXGznpKZ_sA

Here's a visual view of the same idea:

image

▶️ Much the same question has been asked at: https://community.learningequality.org/t/clickable-chapters-on-videos/2707

▶️ Central Question: What is likely the best way to make this critical interactivity (clickable "Video Chapters") happen in a ZIM file, when scraping from YouTube or in any similar way?

ASIDE: YouTube shows the "Video Chapter" name at the bottom &mdash; much like CNN's Chyron ("the crawl").  That more advanced UX/UI is certainly elegant &mdash; but might not be at all necessary (MVP == Minimum Viable Product!)  _A very basic/clickable Table-of-Contents being all that learners (and teachers) need &mdash; to make video-centric ZIM files more practical/pragmatic for everyday learning!_

CONCERN: ZIM files seem to all use video player https://github.com/brion/ogv.js which might make this very difficult, if not impossible?  (i.e. to embed a Clickable-Video-Table-of-Contents within any ZIM file, that students need)  Or am I wrong?!

SIDE QUESTION: Should [warc2zim](https://github.com/openzim/warc2zim) or any other approaches perhaps be considered &mdash; if there's no other/obvious way to make this happen?

arjitcodes:
Fixes #170 
### scrapper
- `yt-dlp` writes a metadata file called `video.info.json` when downloading a video (without an extra yt-dlp call).
- Extract chapters from that metadata file.
- Create a `WEBVTT` file using the chapters.
### zimui
- Check for `chaptersPath` in `video.json`.
- If found, add a new track called chapter with the `WEBVTT` file.

## Chapter in player
Screenshot from 2025-03-12 07-45-37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add support for video chapters (#170)
+
 ## [3.3.0] - 2024-12-05
 
 ### Changed
--- a/scraper/src/youtube2zim/schemas.py
+++ b/scraper/src/youtube2zim/schemas.py
@@ -26,6 +26,14 @@ class Subtitle(CamelModel):
     name: str
 
 
+class Chapter(CamelModel):
+    """Class to serialize data about YouTube Video chapter"""
+
+    start_time: float | int
+    end_time: float | int
+    title: str
+
+
 class Subtitles(CamelModel):
     """Class to serialize data about a list of YouTube video subtitles."""
 
@@ -44,6 +52,8 @@ class Video(CamelModel):
     thumbnail_path: str | None = None
     subtitle_path: str | None = None
     subtitle_list: list[Subtitle]
+    chapters_path: str | None = None
+    chapter_list: list[Chapter]
     duration: str
 
 
--- a/scraper/src/youtube2zim/scraper.py
+++ b/scraper/src/youtube2zim/scraper.py
@@ -58,6 +58,7 @@ from youtube2zim.processing import post_process_video, process_thumbnail
 from youtube2zim.schemas import (
     Author,
     Channel,
+    Chapter,
     Config,
     HomePlaylists,
     Playlist,
@@ -214,6 +215,10 @@ class Youtube2Zim:
     @property
     def subtitles_cache_dir(self):
         return self.cache_dir.joinpath("subtitles")
+
+    @property
+    def chapters_cache_dir(self):
+        return self.cache_dir.joinpath("chapters")
 
     @property
     def videos_dir(self):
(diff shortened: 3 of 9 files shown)


arjitcodes:
@kelson42

arjitcodes:
Thanks for your feedback! The chapters you see in the example you shared are part of a paid plugin. I've implemented a similar effect in our code to enhance chapter visibility on the timeline. Let me know if you have any additional suggestions!
Screenshot from 2025-03-15 18-18-57

benoit74:
Just a small ping to inform you that I did not forgot about this PR but the fact that CI is broken forces me to test this a bit on my own machine, and I would like to test it on a real channel with chapters on some videos, and it takes some time.

benoit74 (changes requested):
There is a bug when you move from a video with chapters (A) to a video without chapters (B) : on video B (and C without chapter, ...), we still have the chapters of video A.

arjitcodes:
I’ve fixed the issue! There were actually two different bugs—one related to timeline modification and another with the Video.js chapter button:

1. First issue: The `addMarker` function was returning early when the chapter list was empty, preventing updates. Removing that check fixed the problem.
3.  Second issue: When switching to a new video, we update the video’s `src` and other options, but (for some reason—maybe because Video.js only adds text tracks during initialization) the text tracks (chapters, subtitles) weren’t updating with the new options.

I first tried using Video.js’s built-in methods like `removeTextTrack` and `addTextTrack`, but they didn’t work in our case. Then, I even attempted manually adding tracks via `appendChild`, but that also didn’t work. So the only solution left (at least for me) was to reinitialize Video.js when switching videos, and that worked!

I actually think reinitializing is a better approach because when we’re not in a playlist and move from one video to another, Video.js already reinitializes. That’s why this issue only occurred when playing videos in a playlist.

Are there any drawbacks that we should consider? WDYT?

benoit74:
Video.JS is indeed a purely JS library, and it is known that these libraries tend to have issues around attributes reloading. The cost of reloading it is moderate and it brings the good advantage that we are sure to free more underlying bugs / weird situations, so I'm fine with this approach.

benoit74 (changes requested):
LGTM ; please squash your commits into one commit, we do not need implementation history (you can have two commits or more if it adds value to split this, but I feel like one is enough, might be wrong), and I will then merge

benoit74 (approved):
LGTM

PR #398, opened 2025-04-02
Add skip controls & keyboard/touch shortcuts

### LINKED ISSUE #397: Add Skip Forward/Backward & Keyboard Shortcuts for Video Player
The video player currently lacks essential playback controls like youtube that enhance the user experience, such as:

1. Skip Forward & Backward – Users should be able to skip forward or backward by a few seconds (e.g., 5s or 10s) like youtube.
3. Keyboard Shortcuts – The player does not support common keyboard shortcuts like: 
  `Space` → Toggle play/pause
  `arrowLeft`  → Skip 5/10 sec forward
  `Ctrl + arrowLeft` → Jump to the next chapter

arjitcodes:
Fix #397 

This PR enhances the video player by adding skip **forward/backward** functionality and **keyboard & touch controls** like youtube for a better user experience.

### Touch Support:

1. `Double-tap left/right` to skip 10s

### Keyboard Shortcuts:

| **Key(s)**       | **Action**                    |
|------------------|--------------------------------|
| `Shift + N`/ `Shift + n`      | Next video                     |
| `Shift + P` / `Shift + p`     | Previous video                 |
| `Arrow Left`     | Skip **-5s**                   |
| `Shift + Arrow Left` | Go to **previous chapter**  |
| `Arrow Right`    | Skip **+5s**                   |
| `Shift + Arrow Right` | Go to **next chapter**     |
| `Arrow Up`       | Increase volume                |
| `Arrow Down`     | Decrease volume                |
| `Space`          | Play/Pause                     |
| `F` / `f` | Toggle fullscreen             |
| `M` / `m` | Toggle mute                   |
| `J` / `j` | Skip **-10s**                 |
| `K` / `k` | Play/Pause                    |
| `L` / `l` | Skip **+10s**                 |
| `>`      | Increase playback speed        |
| `<`      | Decrease playback speed        |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add support for video chapters (#170)
 - Activate links, emails of video description (#390)
+- Add skip forward/backward functionality and keyboard & touch controls (#397)
 
 ## [3.3.0] - 2024-12-05
 
--- a/zimui/src/assets/vjs-youtube.css
+++ b/zimui/src/assets/vjs-youtube.css
@@ -1,5 +1,6 @@
 .vjs-youtube,
 .vjs-youtube video {
+  font-family: 'Roboto', sans-serif;
   border-radius: 8px;
   width: 100%;
   height: auto;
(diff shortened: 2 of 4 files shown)


benoit74:
Some general remarks:
- I really like it, this was definitely missing !
- I think we should document these shortcuts so that users have a chance to find them ; since our user base is mostly offline, I think this documentation should be inside the ZIM, maybe just add adding a question mark which triggers a popup screen detailling these shortcuts is the way to go ; open to other suggestions
- I would rather document `<` and `>` for speed down / up since `Shift+.` and `Shift+,` are linked to the keyboard layout (not all layout have these, mine don't)
- On my Mac, only `>` shortcut works, not `<` ; kinda weird ; but none of these shortcuts work on Youtube so ...
- On my Mac, `Ctrl + arrow` shortcuts do not work at all. Not sure if it can be fixed. Maybe `Shift + arrow` is more intuitive?

arjitcodes:
1. I’ve added a button that opens a shortcuts popup (also accessible via `Shift + ?`).
2. Replaced `Shift + ,`/`.` with `<` and `>` for speed control.
3. Also switched chapter skip from `Ctrl + Arrow` to `Shift + Arrow` for better compatibility.

Just a small note: YouTube actually uses `Ctrl + Arrow` for chapter skipping, not `Shift + Arrow`. But I agree the new mapping feels more intuitive and should work more reliably across devices.

benoit74 (changes requested):
Thank you, works well.

Mostly OK:
- squash all commits into one
- add a CHANGELOG.md entry

benoit74 (approved):
I really like the refactor, way easier to read / maintain, thank you for the extra effort.

PR #437, opened 2025-12-15
change profile and banner flags to blob types

elfkuzco:
## Rationale
See openzim/zimfarm#1576

## Changes
- change profile and banner flags to blob types

### DIFF
--- a/offliner-definition.json
+++ b/offliner-definition.json
@@ -138,7 +138,7 @@
       "description": "Custom profile image. Squared. Will be resized to 100x100px"
     },
     "banner": {
-      "type": "kind",
+      "type": "blob",
       "kind": "image",
       "required": false,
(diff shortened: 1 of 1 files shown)


offliner-definition.json
```diff
@@ -131,13 +131,15 @@
...
-      "type": "url",
+      "type": "blob",
+      "kind": "image",
       "required": false,
       "title": "Profile Image",
       "description": "Custom profile image. Squared. Will be resized to 100x100px"
     },
     "banner": {
-      "type": "url",
+      "type": "kind",
```
benoit74: "type": "blob" should be more appropriate 🤣

benoit74 (changes requested):

benoit74 (approved):
LGTM

PR #447, opened 2026-03-02
Implement dark mode

### LINKED ISSUE #424: Introduce dark theme
This issue is a follow-up of https://github.com/openzim/overview/issues/53

ZIM should AFAP provide a "dark theme" alternative to the standard "light theme" using the CSS `color-scheme` and `prefers-color-scheme` primitives.

This new CSS theme should come either from the upstream (if possible) or should be invented (always if possible).

arjitcodes:
Fix #424 

This PR implements a system-reactive Dark Mode for the ZIMU.

- **System-Reactive Theme**: Integrated a matchMedia listener in loadVuetify for automatic OS-based theme switching.
- **CSS-Only Banner Fallback**: Replaced banner-overlay.png with a dynamic CSS radial-gradient to reduce ZIM size and improve scaling.
- **Video.js UI Fixes**: Used Vue `:deep()` selectors to apply theme-aware styling to dynamically injected shortcut popups.
- **Stability**: Implemented optional chaining in VideoPlayerView and PlaylistPanel to fix Cypress E2E test failures.

>**Why CSS for the Overlay?** 
> Reduces the final ZIM archive size

Screenshot from 2026-03-02 12-42-17
Screenshot from 2026-03-02 12-40-43


zimui/src/components/channel/ChannelHeader.vue
```diff
@@ -22,7 +21,7 @@ onMounted(async () => {
 
 // Computed tabs array based on store data
 const tabs = computed(() => {
-  const baseTabs = [{ id: 0, title: 'Home', to: { name: 'home' } }]
+  const baseTabs = [{ id: 0, title: 'Home', to: { name: 'channel-home' } }]
```
benoit74: Why changing this name? Doesn't look that appealing in browser URL IMHO

zimui/src/components/playlist/panel/PlaylistPanel.vue
```diff
@@ -21,7 +21,7 @@ const props = defineProps<{
   shuffle: boolean
 }>()
 
-const isLoading = computed(() => props.playlist.videos === undefined)
+const isLoading = computed(() => props.playlist?.videos === undefined)
```
benoit74: Why could the prop be null/undefined? (same remark on all changes below)

zimui/src/components/channel/ChannelHeader.vue
benoit74: Would it be possible to fix it the other way around? like keep `home` in the URL and fix any `channel-home` usage?

benoit74 (changes requested):
LGTM, thank you, please add an entry in CHANGELOG.md (with issue number this time) and squash your commits into a single commit.

Issue #448, opened 2026-03-04
Tab underbar slider does not return to the Home tab on navigation


arjitcodes:
**Description**: 
When navigating between channel tabs (e.g., from Home to Videos and back to Home), the Vuetify active tab indicator (the underbar slider) gets stuck and fails to return to the "Home" tab. The route changes correctly, but the UI does not reflect the active state.

**Expected Behavior**:
The tab underbar should correctly highlight the "Home" tab when the user navigates back to the root channel path (/).

**Observed Behavior**:
The active tab indicator remains on the previously selected tab (e.g., Videos).

PR #449, opened 2026-03-04
fix(navigation): resolve home tab underbar not updating

arjitcodes:
Fix: #448 

**What does this PR do?**
Resolves a UI bug where the Vuetify tab slider (underbar) failed to slide back to the "Home" tab when a user navigated back from other tabs like "Videos" or "Shorts".

**Changes included**:
1. Cleaned up `<v-tabs>` template in `ChannelHomeTab.vue` (removed `v-model` and `<router-link>` wrappers).
2. Added `exact` attribute to the `v-tab` loop.
3. Updated `router/index.ts` to assign name: `'home'` to the specific child route (`path: ' '`) rather than the parent route.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


benoit74 (approved):
LGTM, thank you.

Issue #450, opened 2026-03-04
Cypress E2E test crash caused by premature PlaylistPanel rendering


arjitcodes:
**Description:** 
During Cypress E2E tests (`video_player_page.cy.ts`), the test runner occasionally triggers a render before the store has fully resolved the playlist data. This was causing a TypeError: `Cannot read properties of undefined (reading videos)` crash.

PR #451, opened 2026-03-04
Fix: Resolve Cypress E2E crash and prevent premature PlaylistPanel rendering

arjitcodes:
Fix: #450
This PR fixes the intermittent E2E test failures in `video_player_page.cy.ts` by enforcing strict conditional rendering in the parent component (`VideoPlayerView.vue`).


benoit74 (changes requested):
Nice catch, thank you, please add an entry in the CHANGELOG (in same commit, you can force push) with reference to PR number (since we have no GitHub issue here)

Issue #439, opened 2025-12-19
What is the average size of 1-minute ZIM file?


Popolechien:
While dealing with ZIM-requests we often encounter a number of large channels. The question then is whether we should break them down by playlist (or group of playlists), or keep them as a single ZIM file. Since file size is an issue in terms of download and storage for our end users, it would be good to have a basic rule of thumbs that says "1 minute of Youtube create XX GB of ZIM file".

benoit74:
Looks like 2MB/min is a good and easy to remember back-of-the-envelope value to use based on results mentioned in PR.

zeyad-elkholy:
@benoit74 is this issue should still open? the pr is merged so what's still need to be done ?

benoit74:
It is done indeed, thank you.

PR #452, opened 2026-03-09
feat: add a script for 1-min yt zims size

Pranavjeet-Naidu:
tackles #439 

The solution is a python script that: 
- get home_playlists.json from the ZIM to get the list of videos 
- for every playlist and every video in this file, get its slug and duration
- get each video size from the ZIM 
- build a correlation between duration and final webm size ( filesize(MB) = ratio × duration_minutes )
Below is a screenshot attached of the script running locally: 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Resolve home tab underbar not updating (#448)
+
 ### Added
 
 - Implement dark mode (#424)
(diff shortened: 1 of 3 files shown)


analyze_zim.py
```diff
@@ -0,0 +1,93 @@
...
+import re
+import sys
+
+from libzim.reader import Archive
+
+# to parse ISO 8601 duration format
+ISO_DURATION_RE = re.compile(r'^PT(?=\d)(?:(\d+)H)?(?:(\d+)M)?(?:(\d+(?:\.\d+)?)S)?$')
+
+
+def parse_duration(s):
```
benoit74: You should not reinvent the wheel, we already have the `isodate` lib to parse ISO 8601 duration, doesn't this fulfill what we need?

PR #454, opened 2026-03-10
feat: add a script for 1-min yt zims size

Pranavjeet-Naidu:
tackles #439

The solution is a python script that:

get home_playlists.json from the ZIM to get the list of videos
for every playlist and every video in this file, get its slug and duration
get each video size from the ZIM
build a correlation between duration and final webm size ( filesize(MB) = ratio × duration_minutes )

(now includes QA fixes as well)
Below is a screenshot attached of the script running locally:

image


benoit74 (changes requested):
QA is still not passing + the commit misses the CHANGELOG entry we previously introduced

benoit74 (approved):
LGTM, thank you

PR #457, opened 2026-03-15
fix: raise explicit error when zimui/dist is empty or missing 

### LINKED ISSUE #456: Silent broken ZIM when zimui/dist is not built
## Description

When `zimui/dist` has not been built before running the scraper, 
the scraper completes without any error but produces a broken ZIM 
with no homepage.

## Steps to reproduce

1. Clone the repo
2. Set up hatch shell (cd scraper && hatch shell)
3. Run the scraper directly, without first running:
   cd zimui && yarn install && yarn build
4. The scraper runs to completion without any error

## Expected behavior

The scraper should raise an explicit error early, telling the user 
that zimui/dist is missing and what to do about it.

## Actual behavior

The scraper completes silently and produces a broken ZIM. 
The only indication is a cryptic warning in the logs:

> Invalid redirection W/mainPage redirecting to (missing) C/index.html

- Opening the ZIM in Kiwix shows "Page not found" with ERR_FILE_NOT_FOUND
- Opening the ZIM with Docker shows "Page not found" with " The requested path was not found"

## Root cause

Two problems:
1. The README mentions Yarn as a requirement but never instructs 
   the user to run `yarn install && yarn build` in zimui/ before 
   running the scraper
2. The scraper does not check whether zimui/dist is empty or missing 
   before attempting to add its files to the ZIM (scraper.py line ~419)
This bug occurs on first-time setup, before zimui/dist has ever been built. Once built, subsequent runs work correctly

## Proposed fix

1. Update README to document the required build step:
   cd zimui && yarn install && yarn build
2. Add an explicit check in scraper.py before adding zimui files:
   raise ValueError if zimui/dist is empty or missing, with 
   clear instructions on how to fix it

## Environment

- youtube2zim version: youtube2zim 3.5.1-dev0
- Python version: Python 3.10.12
- OS: Ubuntu 22.04

peniel-09:
## Problem

When "zimui/dist" has not been built before running the scraper,
the scraper completes without any error but produces a broken ZIM
with no homepage. The only indication was a cryptic warning:

> Invalid redirection W/mainPage redirecting to (missing) C/index.html

This bug affects first-time contributors who follow the README 
without knowing they need to build the UI first.

## Changes

- Adds an explicit check in `scraper.py` that raises a `ValueError`
  with clear instructions if `zimui/dist` is missing or empty
- Updates README to document the `yarn build` prerequisite

## Testing

Reproduced by deleting "zimui/dist/"and running the scraper.
Before fix: silent broken ZIM produced.
After fix:
 
>  ERROR: zimui/dist is empty or missing. Please build the UI first: 
>   cd zimui && yarn install && yarn build.

Fixes #456

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,27 +21,25 @@ subtitles and the authors' profile pictures; It then produces JSON files contain
 
 For more details / advanced usage than what is in this README, see the [Manual](https://github.com/openzim/youtube/wiki/Manual) and [FAQ/FEE](https://github.com/openzim/youtube/wiki/FAQ---FEE).
 
-# Requirements
+# Prerequisites
 
-
-* [`ffmpeg`](https://ffmpeg.org/) for video transcoding.
-* [`deno`](https://deno.com/) (or another Javascript runtime, see https://github.com/yt-dlp/yt-dlp/issues/15012) for yt-dlp operations when fetching Youtube videos.
-
-* [`Yarn`](https://yarnpkg.com/getting-started/install) to install Javascript dependencies for the Vue.js UI.
-
-**Important:** Before running the scraper, you must build the ZIM UI first:
-
-```bash
-cd zimui && yarn install && yarn build
-```
-
-The scraper will fail with an explicit error if 'zimui/dist' is missing or empty.
+- [Docker](https://docs.docker.com/engine/install/)
+- amd64 architecture
 
 # Installation
 
-Here comes a few different ways to install `youtube2zim`.
+The recommended way to run `youtube2zim` is using the pre-built Docker image:
 
-## Virtualenv
+```sh
+docker pull ghcr.io/openzim/youtube
+```
+
+<details>
+
+<summary>Run from source code</summary>
+
+> [!WARNING]
+> Running from source requires manual setup. Using the Docker image is strongly recommended.
 
 `youtube2zim` is a Python3 software. If you are not using the [Docker](https://docker.com) image,
 you are advised to use it in a [virtualenv](https://virtualenv.pypa.io) to avoid installing software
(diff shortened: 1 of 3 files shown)


benoit74 (changes requested):
Thank you, I like the check in the code, but I'm less convinced about README changes.

Building the UI manually is important only if cloning the repo. Requirements are correct only when not using Docker image.

And Docker image is indeed our preferred way of using the scraper for "new-comers".

I suggest that we enhance further this PR by reworking the README to clearly indicate that Docker is the recommended way of using the scraper + that requirements are only required when using source code. You can get inspiration from https://github.com/openzim/mwoffliner/ where I like the recent changes to push Docker forward.

I wonder if we should fix the wheel to also contain the ZIMUI ; this looks like a bug to me, because it means anyone doing `pip install youtube2zim` gets in fact a broken install

You also need to add an entry to CHANGELOG.md.

README.md
```diff
@@ -21,51 +21,65 @@ subtitles and the authors' profile pictures; It then produces JSON files contain
 
 For more details / advanced usage than what is in this README, see the [Manual](https://github.com/openzim/youtube/wiki/Manual) and [FAQ/FEE](https://github.com/openzim/youtube/wiki/FAQ---FEE).
 
-# Requirements
+# Prerequisites
+
+- [Docker](https://docs.docker.com/engine/install/)
```
benoit74: I think we should also mention `amd64` platform, I don't think the scraper works on `arm64` platforms like Apple Silicon (not tested tbh).

benoit74 (changes requested):
Few more changes, thank you. I will open the separate issue regarding pip packages

benoit74 (changes requested):
LGTM, please squash your commits (or keep only two, one for the check and one for README changes) and rebase on top of main (we do not want to have merge from main to feature branches at openzim).

peniel-09:
> Sorry, there is yet a QA issue to fix : https://github.com/openzim/youtube/actions/runs/23201224871/job/67436012937?pr=457 (and maybe other down the road)

Thanks, fixing it now!

benoit74:
Still QA issues

scraper/src/youtube2zim/scraper.py
```diff
@@ -421,6 +416,23 @@ def run(self):
         logger.info("all done!")
 
     def add_zimui(self):
+        """
+        Fix for zimui dist folder missing or empty
```
benoit74: Please use as much as possible up to the 88 chars limit. Here the text can probably be split in two lines instead of 3

scraper/src/youtube2zim/scraper.py
```diff
@@ -421,6 +416,23 @@ def run(self):
...
+        (which can happen if user forgets to build
+        the UI or if they are using the repo without the UI)
+        """
+        if not (
+            self.zimui_dist.exists()
+            and self.zimui_dist.is_dir()
+            and any(self.zimui_dist.iterdir())
+        ):
+            raise ValueError(
+                "zimui/dist is empty or missing. "
```
benoit74: Same here

benoit74 (changes requested):

scraper/src/youtube2zim/scraper.py
peniel-09: Got it!

benoit74 (approved):
LGTM, thank you!

Issue #432, opened 2025-11-25
[Feature request] Cache more data & add cache expiry


DasDunkel:
Currently only the video and thumbnail are cached, I'd like to propose caching *all* data used by the tool, especially subtitles and chapters, so that each subsequent run would only download metadata/subtitles/etc for newly added videos

This would not only massively speed up re-creating zims when new content is released (90+% of a cached download is waiting on youtube and downloading subtitles), but it'd also reduce the chance of getting rate limited with a high concurrency/large number of zims

And an option to re-download this kind of data if it is older than X amount of time would make this even more useful, so we can maintain a cache while still getting up to date information

benoit74:
Kinda related to https://github.com/openzim/youtube/issues/277, but with a different approach.

We've been stuck so far by the fact that it was not possible to know when these resources would need to be fetched again. I agree that we can't know, the probability to need to update them is low, and it is generally a waste to fetch again over and over. 

I hence tend of like the idea of just having an option to re-download them. Do you think we should be selective with multiple options (re-download videos, re-download subtitles, re-download metadata, ...)? Or have a regex of cached entries to ignore, for more versatility? I'm wondering if we have scenario where we know a specific video subtitles are wrong in the cache and we want to re-download them, but we don't want to waste time / risk being throttled for all other things.

DasDunkel:
I'd probably go for the blanket options (like re-download all subtitles), and add a **include** regex/list, instead of exclude, to make it easier to re-download a small subsection of the data without having to make a regex that'll match everything *except* the ones you want

PR #460, opened 2026-03-17
feat:  Also cache subtitles and chapters in S3 with TTL-based expiry

lumpy72006:
This PR implements the "middle ground" caching strategy discussed with @benoit74 in #432 

## Changes

* A `--subtitles-chapters-cache-expiry-days` CLI flag is added (default: 30) to let operators tune or effectively disable expiry by setting a very large value.

* `download_from_cache` now accepts an optional `max_age_seconds` parameter. When provided, it calls `head_object` on the S3 object to retrieve its `LastModified` timestamp — zero bytes transferred — and returns `False` if the object is older than the threshold, falling back to yt-dlp.

* `download_subtitles` and `generate_chapters_vtt` now check the S3 cache first using this TTL mechanism before falling back to yt-dlp. Fresh data is uploaded to S3 after a successful yt-dlp download so subsequent runs within the expiry window skip yt-dlp entirely.

## Testing

* Tested manually using `moto_server` as a local S3 emulator. On the second run within the same moto session, subtitles were correctly loaded from S3 cache (downloaded ... from cache at subtitles/3p4jR1XcsPQ.json) without hitting yt-dlp. 
Screenshot from 2026-03-17 16-52-36~2
Screenshot from 2026-03-17 16-50-26~2

### DIFF
--- a/scraper/src/youtube2zim/entrypoint.py
+++ b/scraper/src/youtube2zim/entrypoint.py
@@ -222,6 +222,14 @@ def main():
         default=False,
     )
 
+    parser.add_argument(
+        "--subtitles-chapters-cache-expiry-days",
+        help="Number of days a cached file is allowed to live"
+        " before the scraper fetches a fresh copy",
+        type=int,
+        default=30,
+    )
+
     args = parser.parse_args()
     logger.setLevel(logging.DEBUG if args.debug else logging.INFO)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Also cache subtitles and chapters in S3 with TTL-based expiry (#432)
 - Implement dark mode (#424)
 - Added Total playlist duration in the Playlist view and Playlist panel. (#435)
 - Added `analyze_zim.py` contrib script to analyze video duration vs. file size correlation in ZIM files (#439)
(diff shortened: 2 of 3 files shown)


scraper/src/youtube2zim/entrypoint.py
```diff
@@ -218,6 +218,20 @@ def main():
         default=False,
     )
 
+    cache_group = parser.add_argument_group(
```
benoit74: No need for a new group with only 1 setting, especially since we do not use groups elsewhere. If you feel like we should use groups, please open another issue to discuss about that.

benoit74 (changes requested):
See comments

I'm also wondering if we should not delete the files we've downloaded from S3 once used and added to the ZIM. I don't know if this was not handled before this PR either, but I feel like it is useless to keep files on the filesystem for "nothing". Feel free to open another issue if the "flaw" was already present before your PR, I might miss something.

And also please add an entry to CHANGELOG.md

benoit74 (changes requested):
Mostly OK, see my comment and please squash all your commits, so that we have only one commit on top of `main` last commit

benoit74 (approved):
LGTM, thank you

benoit74:
Still QA issues to fix: https://github.com/openzim/youtube/actions/runs/23297595462/job/67777188921?pr=460

benoit74:
Still QA issues : https://github.com/openzim/youtube/actions/runs/23306306207/job/67874096215?pr=460

PR #461, opened 2026-03-18
Bundle Zim UI inside the pip package

### LINKED ISSUE #459: wheel and sdist packages should contain a pre-compiled ZIMUI
Currently when pip package is mostly useless because it does not contains the zimui. This is wrong.

cking100:
Fixes #459 

This approach builds the Zim UI, copies the compiled files into the scraper's  `/templates/dist/` directory, and bundles them with the pip package.

### DIFF
--- a/scraper/pyproject.toml
+++ b/scraper/pyproject.toml
@@ -57,9 +57,6 @@ exclude = ["/.github"]
 
 [tool.hatch.build.targets.wheel]
 packages = ["src/youtube2zim"]
-artifacts = [
-  "src/youtube2zim/templates/assets/**",
-]
 
 [tool.hatch.envs.default]
 features = ["dev"]
@@ -98,7 +95,7 @@ all = "inv checkall --args '{args}'"
 [tool.black]
 line-length = 88
 target-version = ['py314']
-exclude = "(src/youtube2zim/templates/.*|.hatch/.*)"
+exclude = "(src/youtube2zim/zimui/.*|.hatch/.*)"
 
 [tool.ruff]
 target-version = "py314"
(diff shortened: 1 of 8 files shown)


benoit74 (changes requested):
Thank you, well done! 

I kinda like the idea of having the UI `dist` folder as a subfolder of the scraper source code, but I have few struggles:
- instead of manually copying `dist` in the CI, can we modify vite configuration (or anything else) so that this copy is automatically done also for people running `yarn build` manually on their dev machine? Otherwise we will have to update CONTRIBUTING instructions to explain that after running `yarn build` they should copy files, or set environment variable, but I would prefer we keep this only as a last resort, having files placed at the right place directly is more convenient
- please rename `templates` folder to `zimui` ; these are not templates anymore, we should have cleaned this long ago ; we might even get rid of the `dist` subfolder, I'm less sure about this, since it conveys meaning but is mostly useless since we do not have other files there

cking100:
@benoit74 I have made the requested changes as well as fixed the merge conflicts. Please have a look, thank you.

scraper/pyproject.toml
```diff
@@ -58,7 +58,7 @@ exclude = ["/.github"]
 [tool.hatch.build.targets.wheel]
 packages = ["src/youtube2zim"]
 artifacts = [
```
benoit74: Please double check but to me it looks like it is not necessary to specify these artifacts, they are already been included in the sdist + wheel (even if been git-ignored at a higher level, looks like hatchling misses this detail)

benoit74 (changes requested):
Mostly OK, I just realized that we probably do not need the `artifacts` in `pyproject.toml`. 

But this means it is worth to integrate a small check in `run-integration-tests` job to check that Docker image really contains the zimui after installation. Probably a one-liner shell command to run inside the container and check that `zimui` folder is present and contains files.

cking100:
> Please double check but to me it looks like it is not necessary to specify these artifacts

Yeah, it was not needed, I have removed the artifacts.

I have also added a small check in `run-integration-tests` job that runs after docker build to check if zimui contains files inside the container.  @benoit74

benoit74 (approved):
LGTM, thank you very much

PR #464, opened 2026-04-07
Upgrade frontend deps, especially Yarn Modern

benoit74:
Upgrade all frontend dependencies, especially move to Yarn Modern since it is now many years old, and also Vuetify 4 and VueRouter 5.


benoit74:
This is not working as smoothly as I expected, too many things are breaking, especially UI display: many CSS rules have changed, I tried to introduce UnoCSS with Vuetify presets but there are still many bits to tune.

If anyone wants to complete the effort, help is much welcomed

PR #466, opened 2026-04-17
feat:add linux/arm64 support to Docker image

### LINKED ISSUE #458: Check/add support for `arm64` Docker image
We should check if we can provide support for `arm64` Docker builds.

We need to update README prerequisites once done

Abhijithmns:
## Issue

When pulling the Docker image on an `arm64` machine,
users get an `exec format error` because only a `linux/amd64` image is published
to the registry. ARM64 users cannot run the tool at all.

## Test

Tested ARM64 support locally using `docker buildx`:

- Built the image successfully for `linux/arm64`
- Ran the container with a real scrape (`@veritasium`, `--pagination=1`, `--low-quality`)
- Video downloaded, formats merged, and re-encoding started without any architecture errors

All Dockerfile components confirmed ARM64 compatible:
- `node:24-alpine`, `python:3.14-trixie` — multi-arch base images
- `ffmpeg`, `wget`, `aria2` — available on arm64 via apt
- `deno` — official arm64 support via install script

Closes #458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added Total playlist duration in the Playlist view and Playlist panel. (#435)
 - Added `analyze_zim.py` contrib script to analyze video duration vs. file size correlation in ZIM files (#439)
 - Bundle ZIM UI inside pip package (#459)
+- Added `linux/arm64` support to Docker image and CI (#458)
 
 ### Changed
 
--- a/.github/workflows/Publish.yml
+++ b/.github/workflows/Publish.yml
@@ -49,6 +49,7 @@ jobs:
           latest-on-tag: true
           restrict-to: openzim/youtube
           registries: ghcr.io
+          platforms: linux/amd64 linux/arm64
           credentials:
             GHCRIO_USERNAME=${{ secrets.GHCR_USERNAME }}
             GHCRIO_TOKEN=${{ secrets.GHCR_TOKEN }}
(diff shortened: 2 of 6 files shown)


benoit74 (changes requested):
Thank you!

Please:
- ensure we will run all tests in the CI on `arm64` as well, we cannot afford to just hope it will not break in the future: you basically need to duplicate all jobs, and fix existing job names to make it clear they run on `amd64`. Github runner to use is `ubuntu-24.04-arm`
- add an entry in CHANGELOG.md

Abhijithmns:
@benoit74 please check now.
I've made the requested changes:
- Duplicated all CI jobs in Tests.yml and QA.yml to also run on `ubuntu-24.04-arm`
- Added `platforms: linux/amd64 linux/arm64` to both Publish.yml and PublishDockerDevImage.yml
- Added entry in CHANGELOG.md

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added Total playlist duration in the Playlist view and Playlist panel. (#435)
 - Added `analyze_zim.py` contrib script to analyze video duration vs. file size correlation in ZIM files (#439)
 - Bundle ZIM UI inside pip package (#459)
+- Added `linux/arm64` support to Docker image and CI
```
benoit74: Please mention issue number in parentheses

benoit74 (changes requested):
Please fix comment and squash commits into one commit and I will merge

